PR #55, opened 2015-02-10
Feature/template upgrades

joshw:


zenpacklib.py
```diff
@@ -3233,7 +3294,7 @@ def __init__(
             eventClass=None,
             severity=None,
             enabled=None,
-            extra_params=None,
+            extra_params={},
```
cluther: Using a mutable default value for a method argument is dangerous. Especially so with dictionaries. You'll get leakage from one call of the method to the next and end up with surprising default values. There are a couple more cases where this is done with extra_params further down in the code too.

PR #59, opened 2015-02-18
Feature/conversion

joshw:


zenpacklib.py
```diff
@@ -5614,30 +5641,75 @@ def zpl_create(self):
...
+            zenpack = self.dmd.ZenPackManager.packs._getOb(zenpack_name, None)
+            if zenpack is None:
+                LOG.error("Zenpack '%s' not found." % zenpack_name)
+                return
+
+            templates = {}
+            for deviceclass in [x for x in zenpack.packables() if x.meta_type == 'DeviceClass']:
+                dc_name = deviceclass.getOrganizerName()
+                templates[dc_name] = {}
+                for template in deviceclass.getAllRRDTemplates():
```
cluther: Is getAllRRDTemplates() really what you want here? This returns templates in subclasses too which means that there are common cases where the following `templates[dc_name][template.id]` will result in values being overwritten because dc_name/template.id is not unique in the results.

zenpacklib.py
```diff
@@ -5614,30 +5641,75 @@ def zpl_create(self):
...
+                    LOG.error("Diagram type '%s' is not supported.", diagram_type)
             else:
-                print "Usage: %s lint <file.yaml> | py_to_yaml <zenpack name> <__init__.py> | dump_templates <zenpack_name>" % sys.argv[0]
+                print "Usage: %s lint <file.yaml> | py_to_yaml <zenpack name> | dump_templates <zenpack_name> | class_diagram [yuml] <file.yaml>" % sys.argv[0]
+
+
+        def zenpack_templatespecs(self, zenpack_name):
+            zenpack = self.dmd.ZenPackManager.packs._getOb(zenpack_name, None)
+            if zenpack is None:
+                LOG.error("Zenpack '%s' not found." % zenpack_name)
```
cluther: It's either ZenPack or zenpack. Never Zenpack ot henpeck! :)

PR #61, opened 2015-03-03
Restrict relationship infos for ToMany relations.

cluther:
API users were running into serious performance issues because when they
called getInfo on zenpacklib components it would recursively build an
info tree for everything containing under that component.

The IInfo already restricts ToMany relationship information to just show
counts of containing relationships.

This change only impact API consumers. It has no affect on the web
interface.

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -3133,11 +3133,11 @@ class ClassRelationshipSpec(Spec):
         """Return Info properties dict."""
         properties = {}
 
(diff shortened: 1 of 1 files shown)


dougsyer:
chet could this conceivably cause a runaway process on the hub during model  with associated memory leak do you think?

cluther:
@dougsyer: No, this issue is isolated to the API Info adaptor. That code will never be executed by zenhub.

PR #67, opened 2015-03-24
Make !ZenPackSpec root tag implicit.

cluther:
This makes it possible to create a YAML file with no explicit tagging or
typing.


tests/data/zenpacks/ZenPacks.zenoss.ZPLTest1/ZenPacks/zenoss/ZPLTest1/zenpack.yaml
```diff
@@ -1,4 +1,3 @@
-!ZenPackSpec
 name: ZenPacks.zenoss.ZPLTest1
 zProperties: !ZenPackSpec
```
joshw: It can be removed from this line as well, right?

PR #86, opened 2015-04-17
Feature/path control

joshw:


tests/test_install.py
```diff
@@ -98,6 +98,35 @@ def test_upgrade(self):
...
+        p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+                             env=dict(os.environ, ZPL_YAML_FILENAME=yaml_filename))
+        out, err = p.communicate()
+        p.wait()
+        LOG.debug("out=%s, err=%s", out, err)
+
+        self.assertIs(p.returncode, 0,
+                      'Error upgrading %s zenpack: %s' % (self.zenpack_name, err))
+
+        pass
```
cluther: This pass seems unnecessary.

zenpacklib.py
```diff
@@ -806,14 +819,39 @@ def get_facets(self, seen=None):
...
+            if recurse_all:
+                for facet in obj.get_facets(root=root, seen=seen, path=path + [relname], recurse_all=True):
                     yield facet
+                return
+
+            # Otherwise, look at extra_path defined path pattern streams
+            for stream in streams:
+                recurse = any([pattern.match(relpath) for pattern in stream])
+
+                LOG.debug("[%s] matching %s against %s: %s" % (root.meta_type, relpath, [x.pattern for x in stream], recurse))
```
cluther: This seems like useful logging for the ZenPack author, but considering how often our users enable debug logging I think this will be too noisy. Maybe do `LOG.log(9, "msg", ...)` instead? That way people would have to do -v9 or lower to see this level of logging? We've done this before with the WindowsMonitor ZenPack to enable "developer" level logging.

zenpacklib.py
```diff
@@ -1935,15 +1975,19 @@ def __init__(
             :type impacted_by: list(str)
             :param monitoring_templates: TODO
             :type monitoring_templates: list(str)
-            :param filter_display: TODO
+            :param filter_display: Should this class show in any other filter dropdowns?
             :type filter_display: bool
+            :param filter_hide_from: Classes for which this class should not show in the filter dropdown.
```
cluther: This needs to be added to the documentation in docs/yaml-classes-and-relationships.rst.

zenpacklib.py
```diff
@@ -1935,15 +1975,19 @@ def __init__(
...
+            :type filter_hide_from: list(class)
             :param dynamicview_views: TODO
             :type dynamicview_views: list(str)
             :param dynamicview_group: TODO
             :type dynamicview_group: str
             :param dynamicview_relations: TODO
             :type dynamicview_relations: dict
-            # TODO: should make this a spec class, not a plain dict.
+            :param extra_paths: TODO
```
cluther: `ClassSpec.extra_paths` needs to be added to the documentation in docs/yaml-classes-and-relationships.rst.

zenpacklib.py
```diff
@@ -6133,6 +6266,57 @@ def zpl_create(self):
                 else:
                     LOG.error("Diagram type '%s' is not supported.", diagram_type)
 
+            elif len(args) == 2 and args[0] == "list_paths":
```
cluther: `list_paths` command needs to be documented in docs/command-line-reference.rst.

tests/test_install.py
joshw: This was added to reproduce a bug that eric had reported (now fixed).  Basically, in-place upgrade where you add a monitoring template.

zenpacklib.py
joshw: I had that concern too.  I didn't know about the "9" trick, but I will do that.

PR #97, opened 2015-06-05
Set rrdtype, default GAUGE

pcarinhas:


cluther:
Opened #99 to get this fix merged into hotfix/1.0.2 instead of develop.

PR #100, opened 2015-06-10
Fixes ZEN-18269

j053ph4:
Revised the get_catalog_scopes method to return only the specified
scope(s)

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -441,13 +441,14 @@ class CatalogBase(object):
         """Return catalog scopes by name."""
         spec = cls._get_catalog_spec(name)
         if not spec:
-            []
-
-        scopes = [spec['indexes'][x].get('scope', 'device') for x in spec['indexes']]
-        if 'both' in scopes:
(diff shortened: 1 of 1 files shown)


cluther:
I think there's more to that issue than the fix in the pull request addresses.

The problem seems to be that get_catalog_scope's scope is a component+catalog, and not component+catalog+index. So if one index in the catalog has a global scope and another has a device scope get_catalog_scopes returns {'device', 'global'} which in turn creates the catalog in both scopes. So the reason all global catalogs also get created as device catalogs is that there's a default 'id' index in all catalogs that has a 'device' scope. I'd think that catalogs that had all indexes with a 'device' scope would only be created for the device, not globally. So right now 'global' has the same meaning as 'both'.

Your fix just ignores indexes that haven't explicitely set their scope. Here's where that default 'device' scoped 'id' index gets added to all catalogs.

https://github.com/zenoss/zenpacklib/blob/develop/zenpacklib.py#L2349

dougsyer:
yeah thats what i was seeing on the hp zenpack, global = both.

PR #112, opened 2015-08-19
Adding .gitignore file

jstanley23:
Adding a default .gitignore file to ZenPack base directory. Since GitHub is the default supported repository for storing ZenPacks for the community.


cluther:
Good idea. Holding on merging until I figure out if I'm allowed to.

vikt0rs:
Please, add a `.tox/.` directory to the `.gitignore` file

j053ph4:
This is a very old PR not sure if it's still valid (i think .gitignore is already there?)

j053ph4:
yeah this is no longer valid...

Also this applies to the '--create' zp arg, which isn't necessarily the same thing as adding a Git repo

PR #114, opened 2015-08-27
Fix inherited rels and props not showing in UI.

cluther:
Directly inherited relationships and properties were already working,
but not from any higher in the inheritance tree.


dougsyer:
Thanks i thought i was loosing it..

PR #136, opened 2016-01-11
Use Acquisition to get dc

pcarinhas:
Fixes: ZEN-18824

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -6433,8 +6433,9 @@ if __name__ == '__main__':
                 return
 
             templates = collections.defaultdict(dict)
-            for deviceclass in [x for x in zenpack.packables() if x.meta_type == 'DeviceClass']:
-                for template in deviceclass.getAllRRDTemplates():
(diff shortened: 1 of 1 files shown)


zenpacklib.py
```diff
@@ -6422,8 +6422,9 @@ def zenpack_templatespecs(self, zenpack_name):
                 return
 
             templates = collections.defaultdict(dict)
-            for deviceclass in [x for x in zenpack.packables() if x.meta_type == 'DeviceClass']:
-                for template in deviceclass.getAllRRDTemplates():
+            for ac_wrapper in [x for x in zenpack.packables() if x.meta_type == 'RRDTemplate']:
```
cluther: I think this fix introduces some new problems.

1. ZenPacks that have a device class in their packables, but no monitoring templates explicitely won't dump the monitoring templates, or the device class.
2. ZenPacks that have monitoring templates explicitely in their packables, but not the containing device class will have the device class dumped along with all monitoring templates in the device class even if they shouldn't be part of the ZenPack.

cluther:
Merged this into 1.0.8 with db35e8fdb4a222657b725f13455c9e5715f652a0.

PR #137, opened 2016-01-20
Fix dynamicview_relations type issue.

cluther:
When zenpacklib created BaseRelation objects for DynamicView for defined
dynamicview_relations, it would pass in the type/tag as a string instead
of a iterable of strings. This would end up tagging the relationship
with one tag per character in the string instead of the correct name.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,11 @@
+Release 1.0.7
+=============
+
+Defects fixed
+-------------
+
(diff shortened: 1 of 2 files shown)


cluther:
This issue was found when converting the Ceph ZenPack from using "impacts" and "impacted_by" to using "dynamicview_relations".

PR #138, opened 2016-01-21
Fix filtering of monitoring templates out of exported XML.

joshw:
I'm a little baffled at this.  I know that when I wrote it, I had to do the explicit acquisition wrapper stuff to make sure that the wrapped objects were both zope proxies and kept their acquisition behavior.  It seems like at some point something changed, and now they are actually keeping their acquisition, but losing their proxying when I reapply the acquisition. 

I can't really explain why or what could be different, so i've tried to write this somewhat defensively so it will do whatever it needs to do, or log errors if it finds something wrong such that the export may not work properly.

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -49,6 +49,8 @@ import sys
 import math
 import types
 
+from lxml import etree
+
 if __name__ == '__main__':
     import Globals
     from Products.ZenUtils.Utils import unused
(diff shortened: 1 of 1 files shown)


dougsyer:
Thanks!

cluther:
I tested this by creating a ZenPack with a mixture of device classes and monitoring templates. Some associated with the ZenPack in the UI, and some supplied by zenpack.yaml. With the current develop branch zenpacklib.py, I found that exporting the ZenPack would erroneously include yaml-supplied monitoring templates that were in UI-associated device classes. I believe this is the issue trying to be solved here. Correct me if I'm wrong.

Attempting an export with this branch's version of zenpacklib.py resulted pretty much the entirety of dmd to be exported into objects.xml except for most monitoring templates.

joshw:
Fixed the issues we identified.  Let's try again.

joshw:
OK, trying again....

joshw:
bump

joshw:
This is being tracked in ZEN-21697

dougsyer:
One other thing i noticed is the templates report zenpack ignores zenpacklib templates.

cluther:
@dougsyer: That's because currently the YAML templates aren't being added to the ZenPack's packables relationship. That's going to be corrected too.

cluther:
Merged into hotfix/1.0.9 in #141.

PR #140, opened 2016-01-26
Fix non-containing setters with standard devices.

cluther:
Fixes ZEN-21747.

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -716,7 +716,7 @@ class ComponentBase(ModelBase):
             return
 
         # Find and add new object to relationship.
-        for result in self.device().search('ComponentBase', id=id_):
+        for result in catalog_search(self.device(), 'ComponentBase', id=id_):
             new_obj = result.getObject()
             relationship.addRelation(new_obj)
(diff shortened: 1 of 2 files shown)


zenpacklib.py
```diff
@@ -760,7 +760,7 @@ def setIdsInRelationship(self, relationship, ids):
         query = Or(*[Eq('id', x) for x in changed_ids])
 
         obj_map = {}
-        for result in self.device().search('ComponentBase', query):
+        for result in catalog_search(self.device(), 'ComponentBase', query):
             obj_map[result.id] = result.getObject()
```
eimajenthat: Looks like catalog_search handles cases where the device doesn't have a search method defined??

eimajenthat:
This PR is pointed at the Hotfix branch, not develop.  Just want to double check that's the desired outcome before merging.

zenpacklib.py
dbouchillon: sorta.  it handles cases where a zpl component is using the base device class instead of the zpl device

PR #145, opened 2016-02-18
Feature/container relationships

j053ph4:
Added method to handle custom relation names for container components (higher-degree ancestors)

These are otherwise displayed improperly within the component grid and the details pane


j053ph4:
Bumping for review.

Currently this is needed for Azure and BigIpMonitor ZP for correct display of parent container relations

PR #147, opened 2016-03-09
fixes ZEN-21669: Dynamic View in Display drop-down displays only comp…

itshane:
…onents with relations.


cluther:
I've opened a zenpacklib-specific issue (ZEN-22391) and pull request #149 against it. The pull request is essentially the change in this pull request with a couple of differences.

* Using "service_view" in dynamicview_views as the test instead of the existence of any dynamicview_relations.
* Added a unit test.
* Documented in CHANGES.
* Pull request against 1.0.11 hotfix instead of develop.

Would you review #149?

PR #148, opened 2016-03-09
Initialize CFG in __init__.py for TestCase

cluther:
zenpacklib.TestCase.afterSetUp() assumes that the ZenPack module will
have a CFG attribute with a value of the ZenPack's ZenPackSpec. The
problem was that running "zenpacklib.py create ..." didn't create
__init__.py properly to make this happen.

This change creates __init__.py in a way that will allow
zenpacklib.TestCase to work.

Fixes ZEN-22387.


cluther:
Bump for review.

PR #150, opened 2016-03-29
Guard all getObject() calls agaist damaged ZoDB

pcarinhas:
Fixes ZEN-22592

Wrap all getObject() calls in try, provide log in case of exception.


cluther:
Cherry-picked into hotfix/1.0.12 as 889ca503b13455677151076e493a75ee73cb5983.

PR #153, opened 2016-04-20
Fixes modeler timeout issue

j053ph4:
Affects ZPL-based Zenpacks using indexed component attributes.  In
certain cases the modeler times out due to the following reasons:

1) index scope too wide for catlog objects since ZPL will always choose
'Products.ZenModel.DeviceComponent.DeviceComponent' objects in the
catalog (all components) instead of just objects of the relevant
component class
2) presence of bad catalog brain objects will cause modeler tracebacks
if getObject() fails, which also causes modeler to fail.

- Fixed by narrowing focus of catalog results scope to relevant class
entries
- Added commented code to uncatalog bad entries if appropriate to do so
here

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -624,17 +624,21 @@ class CatalogBase(object):
                 # reindex all objects of this type so they are added to the
                 # catalog.
                 results = ICatalogTool(context).search(types=(classname,))
+
+                # global catalog (I think) for removing bad entries
+                #zcat = context.getDmd().global_catalog
+
                 for result in results:
                     try:
                         ob = result.getObject()
                         if hasattr(ob, 'index_object'):
                             ob.index_object()
                     except:
-                        LOG.ERROR('Problem indexing %s' % result.getPath())
+                        LOG.ERROR('Problem indexing bad catalog entry: %s' % result.getPath())
                         # not sure if this is appropriate to do here or not.
                         # but these bad paths didn't show up in dmd.global_catalog
                         # without using ICatalogTool
-                        zcatalog.uncatalog_object(result.getPath())
+                        # zcat.uncatalog_object(result.getPath())
 
     def index_object(self, idxs=None):
         """Index in all configured catalogs."""
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -584,7 +584,7 @@ class CatalogBase(object):
         return zcatalog
 
     @classmethod
-    def _create_indexes(cls, zcatalog, spec):
+    def _create_indexes(cls, zcatalog, spec, leaveObjects=True):
         from Products.ZCatalog.Catalog import CatalogError
         from Products.Zuul.interfaces import ICatalogTool
         catalog = zcatalog._catalog
@@ -592,7 +592,7 @@ class CatalogBase(object):
         # I think this is the original intent for setting classname, not sure why it would fail
         try:
             classname = '%s.%s' % (cls.__module__, cls.__class__.__name__)
-        except:
+        except Exception:
             classname = 'Products.ZenModel.DeviceComponent.DeviceComponent'
 
         for propname, propdata in spec['indexes'].items():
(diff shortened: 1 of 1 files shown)


dougsyer:
Wil this cause a race condition on a hub?  Ive seen that behavior

j053ph4:
Hi Doug, I don't think it's a race condition exactly, but, as currently structured, all "DeviceComponent" objects in the global catalog get polled and reindexed (if specified), rather than just objects of the relevant catalog/component class type.  This is behind the timeouts (and I would guess high memory utilization) when modeling, and would affect any ZenPacks with indexed component attributes (including the latest releases of HP Proliant, AixMonitor, and IBM Power, but those are just the ones I know of).

Also, the presence of bad catalog objects (where "getObject" fails) causes a traceback which also prevents the modeler from finishing or the component catalog from getting created.

zenpacklib.py
```diff
@@ -621,9 +624,21 @@ def _create_indexes(cls, zcatalog, spec):
                 # reindex all objects of this type so they are added to the
                 # catalog.
                 results = ICatalogTool(context).search(types=(classname,))
+
```
dbouchillon: there's no need to keep commented out code in

zenpacklib.py
```diff
@@ -621,9 +624,21 @@ def _create_indexes(cls, zcatalog, spec):
...
+                #zcat = context.getDmd().global_catalog
+
                 for result in results:
-                    if hasattr(result.getObject(), 'index_object'):
-                        result.getObject().index_object()
+                    try:
+                        ob = result.getObject()
+                        if hasattr(ob, 'index_object'):
+                            ob.index_object()
+                    except:
```
dbouchillon: you need to catch something here, like `Exception`

zenpacklib.py
```diff
@@ -589,8 +589,11 @@ def _create_indexes(cls, zcatalog, spec):
         from Products.Zuul.interfaces import ICatalogTool
         catalog = zcatalog._catalog
 
-        classname = spec.get(
-            'class', 'Products.ZenModel.DeviceComponent.DeviceComponent')
+        # I think this is the original intent for setting classname, not sure why it would fail
+        try:
+            classname = '%s.%s' % (cls.__module__, cls.__class__.__name__)
+        except:
```
dbouchillon: Also `except Exception` here

zenpacklib.py
j053ph4: Dave, I left that commented code in because I wanted to see whether the catalog retrieval failure should be dealt with then and there (by uncataloging) or just reported for later action.  What do you think?

I could also uncomment it, but add a method argument like "leaveObjects" defaulted to False (reporting only)

joshw: There is actually another pull request open to do something similar (https://github.com/zenoss/zenpacklib/pull/150)  - Can you merge those other guards into this PR?

zenpacklib.py
```diff
@@ -584,13 +584,16 @@ def _create_catalog(self, name, scope='device'):
...
+    def _create_indexes(cls, zcatalog, spec, leaveObjects=True):
         from Products.ZCatalog.Catalog import CatalogError
         from Products.Zuul.interfaces import ICatalogTool
         catalog = zcatalog._catalog
 
-        classname = spec.get(
-            'class', 'Products.ZenModel.DeviceComponent.DeviceComponent')
+        # I think this is the original intent for setting classname, not sure why it would fail
+        try:
+            classname = '%s.%s' % (cls.__module__, cls.__class__.__name__)
```
joshw: I believe this is a reasonable default, however, I can't help but think that there was some reason for the DeviceComponent thing.  I don't think it has to do with subclassing as I think the base class would suffice here, but perhaps it has to do with monkeypatching or something? 

All of this indexing code originally was derived from a mixin that Chet used to use.  Before we change this, i'd just like to see if he has any ideas as to why it was that way.   To me, though, this looks fine.  Though I wouldn't bother with the try block and fallback- I would expect this to always succeed, unless cls isn't a class- at that point, I think we have bigger problems.

zenpacklib.py
```diff
@@ -621,9 +624,23 @@ def _create_indexes(cls, zcatalog, spec):
...
+                        if hasattr(ob, 'index_object'):
+                            ob.index_object()
+                    except Exception:
+                        LOG.error('Problem indexing bad catalog entry: %s' % result.getPath())
+                        # not sure if this is appropriate to do here or not.
+                        # but these bad paths didn't show up in dmd.global_catalog
+                        # without using ICatalogTool
+                        if not leaveObjects:
+                            LOG.info('Removing bad catalog entry: %s' % result.getPath())
+                            zcat.uncatalog_object(result.getPath())
```
joshw: I vote for not doing this at all, even with the option.  We don't generally do it elsewhere in our code- I think the assumption is that if your catalog is showing stale entries, we want you to get out the toolbox and scan it properly.

zenpacklib.py
j053ph4: Agreed -- I wrote it this way because like I said I'm guessing at the original intent

zenpacklib.py
j053ph4: I added the guards from https://github.com/zenoss/zenpacklib/pull/150 but I'm not sure if I should have just removed the guards on my side and left them in the other PR?

PR #162, opened 2016-05-27
Add dynamicview_weight field to classes

cluther:
* Fix documentation for dynamicview_group class field
* Allow specifying absolute path in icon class field


dougsyer:
Awesome i was hoping for this..can now have my component sort order in the tree alphabetical and sort my dynamic view more sensibly..

PR #164, opened 2016-06-03
Fix bug where we attempt to set values on zproperties before they hav…

joshw:
…e been created, during zenpack installation.

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -197,6 +197,8 @@ class ZenPack(ZenPackBase):
             d.buildRelations()
 
     def install(self, app):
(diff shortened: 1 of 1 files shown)


dougsyer:
There is also a long standing issue where its a giant pain to change a zproperty category but thats probably a core issue

cluther:
@dougsyer: Are you referring to what appears in the category column when you're looking at the configuration properties? I just checked and changing the value of the zProperty's category field in zenpack.yaml does change that value after only restarting Zope.

It's not even necessary to reinstall the ZenPack to change the category because it's something that's registered in the code, not in ZODB.

PR #174, opened 2016-06-16
Device status is now critical event(s) in /Status

cluther:
This seems like the simplest way to think of status that applies to as
many different types of devices as possible. If you want to mark your
device as "down", create a critical severity event for the device in, or
under, the /Status event class.

Devices that are not being monitored because their production state is
less than zProdStateThreshold will still have an unknown status.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,7 @@ Features
 * Add overridable getDynamicViewGroup method to generated classes.
 * Class icons beginning with / will be treated as absolute URL paths.
 * Improve performance of entity properties in component grids.
+* Simplify what device status means to critical event(s) in /Status.
 
 Documentation
 
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -114,6 +114,12 @@ from Products.Zuul.utils import ZuulMessageFactory as _t
 from zope.publisher.interfaces.browser import IDefaultBrowserLayer
 from zope.viewlet.interfaces import IViewlet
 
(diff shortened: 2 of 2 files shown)


dougsyer:
If im not mistaken though in order for the daemon you are using to stop polling the device you need a /Starus/Ping event > informational..unless you use enterprise and the snmp daemon uses the localping manager..so this is really just l an effect in the gui im assuming..

dougsyer:
Philosophically inwant to know a) is our ability to monitor the device impared and b) if not, is it healthy.

In the case of a) you usually cant assumena device is down, it might as well be down if you cant monitor it but you dont really know a device is defraded unless it tells you it is..or you can poll a status somehow..

In eirher case a) or b) i may want to trigger an action to stop polling the device until the condition clears..

cluther:
This change is really just a UI change. The event filter used by the collector framework to determine which devices shouldn't even be monitored is this..

- eventClass in [/Status/Ping]
- severity in [warning, error, critical]
- status in [new, acknowledged, suppressed]

Specifically it ends up being ZepFacade.getDevicePingIssues() that uses this filter to let the collector daemons know which devices are not responding to pings.

We realize that a lack of response from a device on any protocol doesn't necessarily indicate it's down. If we're getting really pedantic, down doesn't even mean "critically broken" or really have any specific meaning related to a computer system. We're just going with conventional terminology, with the point being that someone should probably take a look at this thing.

PR #177, opened 2016-07-07
Fixes ZEN-19798: Honors display property

alam-zenoss:


cluther:
@yichi-lu, @alam-zenoss: A few notes on this pull request even though it's already been merged.

1. While the fix makes things better than they were before, it results in the general _display_ option overriding values set for _details_display_ and _grid_display. It seems like the more specific options should win if both the general and specific are set.

2. Since this is a bug fix and introduces no new functionality it should go into a 1.0.x hotfix instead of develop.

3. There's no update to the CHANGES file documenting this fix.

cluther:
Scratch note #1 from above. I discussed it with @alam-zenoss, and he convinced me that this was better.

PR #178, opened 2016-07-07
Removed unexpected keyword 'default'

alam-zenoss:


cluther:
@yichi-lu, @alam-zenoss: A couple of notes on this even though it's already been merged.

1. As a bug fix that introduces no new functionality this should have gone into a hotfix branch instead of develop.

2. There's no update in the CHANGES file documenting this fix.

PR #179, opened 2016-07-12
getRRDTemplateName can return label of base class (Fixes ZEN-19025)

j053ph4:
- pull last value from _templates in getRRDTemplateName instead of first
value, since this
value is the desired default template in getRRDTemplate method


j053ph4:
bump please

PR #184, opened 2016-07-15
Ensure non-cached datapoints return current value

j053ph4:
- Fixes ZEN-22288

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -49,6 +49,7 @@ import re
 import sys
 import math
 import types
+import time
 
 from lxml import etree
 
@@ -2552,7 +2553,7 @@ class ClassSpec(Spec):
                     if cached:
                         r = self.cacheRRDValue(datapoint, default=default)
                     else:
-                        r = self.getRRDValue(datapoint)
+                        r = self.getRRDValue(datapoint, start=time.time()-1800)
 
                     if r is not None:
                         if not math.isnan(float(r)):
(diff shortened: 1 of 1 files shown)


cluther:
@j053ph4: What about metrics that are collected every 15 minutes or less frequently?

j053ph4:
The 1800 value should get everything within the last 30 mins, but should there be some function to attempt to determine the cycletime of the data point?  There's probably some easy way to do that w/the RRD, but not sure about OpenTSDB (or if this problem even applies to OpenTSDB)?

cluther:
The issue in ZEN-22288 is specific to Zenoss 4, and it's affected by the RRA configurations of the datapoint which may, or may not, be affected by the datasource's cycle time.

@iwills-zenoss found that by default, getRRDValue was looking for values with a start time of 36 hours ago, and an end time of now. Our default RRA configuration only allows for storage of 600 samples in its unaggregated RRAs. That would be 50 hours for a datapoint with a 5 minute collection interval, and it getRRDValue would work. However, 600 samples at a 1 minute collection interval can only hold 10 hours of data and getRRDValue would be forced to look into one of the aggregating RRAs.

cluther:
30 minutes is probably a safe assumption for a maximum cycletime, but I'd like to see the start time only set on Zenoss 4 since there's no reason to set it on Zenoss 5.

A note should also go into CHANGES about this change. It also seems like a bug fix rather than any new functionality, so the change should target a hotfix branch, not develop.

PR #185, opened 2016-07-19
Ensure non-cached datapoints return current value

j053ph4:
- Fixes ZEN-22288
- adds platform version detection function
- "start" parameter used in applies to Zenoss 4.x and below only

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -2448,8 +2448,10 @@ class ClassSpec(Spec):
                     if cached:
                         r = self.cacheRRDValue(datapoint, default=default)
                     else:
-                        r = self.getRRDValue(datapoint, default=default)
-
+                        if is_pre_zenoss_5():
+                            r = self.getRRDValue(datapoint, start=time.time()-1800)
+                        else:
+                            r = self.getRRDValue(datapoint)
                     if r is not None:
                         if not math.isnan(float(r)):
                             return r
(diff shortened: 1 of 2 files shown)


zenpacklib.py
```diff
@@ -2448,8 +2448,10 @@ def datapoint_method(self, default=spec.datapoint_default, cached=spec.datapoint
                     if cached:
                         r = self.cacheRRDValue(datapoint, default=default)
                     else:
-                        r = self.getRRDValue(datapoint, default=default)
-
+                        if is_pre_zenoss_5():
```
cluther: It seems unnecessarily expensive to make this check every time getRRDValue needs to be called when its result is always the same thing. I'd recommend setting a zenpacklib module global that figures this out once so we're only checking its value here.

Something like this (use feature checking, not explicit version checking) in the zenpacklib module scope.

```python
try:
    from Products.Zuul.facades import metricfacade  # noqa: imported only to test that it can be imported
except ImportError:
    HAS_METRICFACADE = False
else:
    HAS_METRICFACADE = True
```

Then switch this section to..

```python
if HAS_METRICFACADE:
    r = self.getRRDValue(datapoint)
else:
    r = self.getRRDValue(datapoint, start=time.time()-1800)
```

cluther:
You need to add a note to the _Fixes_ section of _Release 1.0.14_ in _CHANGES_.

j053ph4:
bump please

PR #186, opened 2016-07-19
prevent relationship schema from overwriting if duplicate names used

j053ph4:
- fixes bug introduced by PR 182 (ZEN-23763)
- HP Proliant relations are an example case

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -1625,7 +1625,8 @@ class ZenPackSpec(Spec):
 
                     # make sure we have the schema defined
                     if target_relname in class_.relationships:
-                        class_.relationships[target_relname].schema = target_schema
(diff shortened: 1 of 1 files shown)


zenpacklib.py
```diff
@@ -1625,7 +1625,8 @@ def get_bases(cls, bases=[]):
 
                     # make sure we have the schema defined
                     if target_relname in class_.relationships:
-                        class_.relationships[target_relname].schema = target_schema
+                         if not class_.relationships[target_relname].schema:
+                             class_.relationships[target_relname].schema = target_schema
```
pcarinhas: zenpacklib.py|1628 col 26 error| indentation is not a multiple of four [E111]                      
zenpacklib.py|1629 col 30 error| indentation is not a multiple of four [E111]

PR #188, opened 2016-07-21
list_paths improvements

joshw:
Avoid doubling back during recurse_all traversal
Take containment into account as well, since this is also figured into
the dropdown in the UI.  Now shows a more accurate list of what will
be reachable from what.

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -928,7 +928,7 @@ class ComponentBase(ModelBase):
 
         return faceting_relnames
 
-    def get_facets(self, root=None, streams=None, seen=None, path=None, recurse_all=False):
+    def get_facets(self, root=None, streams=None, seen=None, path=None, skip=None, recurse_all=False):
         """Generate non-containing related objects for faceting."""
         if seen is None:
             seen = set()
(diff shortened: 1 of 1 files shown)


cluther:
Is this fixing a bug? If so...

- can we get a JIRA that describes the bug?
- and get this PR targeting the hotfix branch instead?
- and a test that fails without this change?

Regardless of whether it's a bugfix or a change in functionality..

- could we get a corresponding update to CHANGES?

joshw:
It is fixing two issues that caused list_paths's output to be incomplete, yes.  I was encountering a version of ZEN-23840, and a its fix wasn't complete for my use case, so I further improved it.

I then observed that list_paths wasn't showing all the reachability that I expected to, because it didn't take containment into account, so I added that.  That is either a bug fix or an enhancement, depending on your expectations.   I can open a jira on that if you would like.

I don't know that i can write a non-brittle test for this- i'd basically be looking for specific output on list_paths with a sample zenpack.  I could do that, but i'm not sure that it's worth the headache of having it break later if we change the output.  If you'd like it, I can do so.

Regarding hotfix vs. develop, I could do that if you want, I guess my main concern is getting it into whatever version of ZPL I will be using in vsphere, for the next few days at least.   I believe due to other bugfixes that are ongoing, it's best for me to be on develop for the moment, so i'd like to get this fix over there too.   I expect to update to a released version as soon as I finish the conversion, and once one is available that includes all the fixes I end up needing.

cluther:
So it sounds like it's fixing a bug to me. That'd mean all of the bullet points above would apply. It's OK for unit tests to break if they're testing output that changes. I'd hope it would break in that case.

As for getting it into develop, you can merge the hotfix branch into develop as soon as this gets into it. The hotfix branch can be merged into develop anytime.

PR #190, opened 2016-07-22
fix issue with lost relations and wrongly assigned schemas (ZenPack)

j053ph4:


j053ph4:
merge after:
https://github.com/zenoss/zenpacklib/pull/189

PR #191, opened 2016-07-22
list_paths improvements (ZenPack)

j053ph4:
Avoid doubling back during recurse_all traversal
Take containment into account as well, since this is also figured into
the dropdown in the UI. Now shows a more accurate list of what will
be reachable from what.


j053ph4:
should be merged after:

https://github.com/zenoss/zenpacklib/pull/188

PR #192, opened 2016-07-22
getRRDTemplateName can return label of base class (Fixes ZEN-19025 in ZenPack)

j053ph4:


j053ph4:
merge after merge of
https://github.com/zenoss/zenpacklib/pull/179

PR #193, opened 2016-07-22
Ensure non-cached datapoints return current value (ZenPack)

j053ph4:
- Fixes ZEN-22288
- "start" parameter used in applies to Zenoss 4.x and below only


j053ph4:
merge after:

https://github.com/zenoss/zenpacklib/pull/185

j053ph4:
merge after:
https://github.com/zenoss/zenpacklib/pull/185

PR #194, opened 2016-07-22
Support for using enum property with datapoints (ZenPack)

j053ph4:
- allow mapping datapoint output to enum strings


j053ph4:
merge after:
https://github.com/zenoss/zenpacklib/pull/181

PR #195, opened 2016-07-22
Incorrect display of nested custom-named relations (ZenPack)

j053ph4:
- Fixes ZEN-23995
- Added method to handle custom relation names for nested components
- These are otherwise displayed improperly within the component grid and
the details pane


j053ph4:
merge after:

https://github.com/zenoss/zenpacklib/pull/180

PR #201, opened 2016-07-26
Check for reserved keywords from python, Device, Component

dbouchillon:
Fixes ZEN-19460

This will load in methods/properties from Device, Component, and
python reserved keywords and test for them.  Added another list
for any future keywords for which we should look.  Unit test checks
one yaml that is properly named and one that is using reserved words.

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -167,9 +167,18 @@ TestCase = None
 # Required for registering ZCSA adapters.
 GSM = getGlobalSiteManager()
 
-# Use OTHERS for any other words we shouldn't use
-OTHERS = ['uuid', ]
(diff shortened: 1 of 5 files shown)


zenpacklib.py
```diff
@@ -166,6 +167,10 @@
 # Required for registering ZCSA adapters.
 GSM = getGlobalSiteManager()
 
+# Use OTHERS for any other words we shouldn't use
+OTHERS = ['uuid', ]
+KEYWORDS = keyword.kwlist + list(set(dir(BaseDevice)).union(set(dir(BaseDeviceComponent)))) + OTHERS
```
joshw: You may also want to add the ones from (Base)Device/ComponentInfo

PR #217, opened 2016-08-03
consistent use of OrderedDict class

j053ph4:
there looks to be a safeguarded import of collections.OrderedDict vs. a
backported version (guessing that would be for python 2.6)

Whether the existence of the preferred collections.OrderedDict is still
a concern or not, some zenpacklib code refers to the "safe" OrderedDict
while other parts use collections.OrderedDict directly.

I imagine that the use of the 2 different classes could possibly cause
issues, so I propose changing all references to collections.OrderedDict
to the locally defined version.  

Alternately, I'd propose we do away with the local class and refer
exclusively to the collections version


pcarinhas:
I'd second your proposal to use the collections (standard) version of OrderedDict for forward compatiblity,
and sanity.

cluther:
I agree with removing the conditional detection and assume Python 2.7.

j053ph4:
OK cool -- I'll update this branch to remove the custom OrderedDict and to assume python 2.7 and collections exist

j053ph4:
OK I updated the PR to remove the conditional detection and assume Python 2.7

PR #219, opened 2016-08-09
We should only look for callable attributes and some js words

dbouchillon:
Fixes ZEN-19460

Changed so that we look at only callable attributes and included
some js words like uuid


dbouchillon:
bump

dbouchillon:
bump

PR #220, opened 2016-08-11
Multiple YAML file support, support for optimized YAML formatting

j053ph4:
- added capability to build a ZenPackSpec from multiple YAML files
(load_yaml_multi)
- build ZenPackSpec from YAML files in a directory (load_yaml_dir)
- optimize existing YAML file format and DEFAULTS (optimize_yaml)
- added -o (--optimize) switch to zenpacklib.py options
- added "compare_zenpackspec" to report whether different YAML documents
result in the same ZenPackSpec
- added unit tests "test_optimized_yaml" and "test_load_yaml_multi"
- refactor of Loader and Dumper classes
- converted all functions with "loader" or "dumper" parameters into
Loader/Dumper class functions
- added Dumper representer for Python Unicode and OrderedDict
- added Loader constructor for OrderedDict
- better order preservation when dumping ZenPackSpec to YAML (not 100%)
- removed extraneous "!!ZenPackSpec" and !!unicode annotations from
dumped YAML
- added 'type_map' method that attempts to determine python data type
(ZEN-24079)
- included python data type checking for extraparameters in
"construct_spec" (ZEN-24079)
- added zen-27049-extraparams.yaml for unit test validation
- added alphabetical sort order when building specs in
"specs_from_param", since this was tripping up inter-ZenPackSpec
comparison (OrderedDicts with different orders)
- removed unneeded imports from zenpacklib.py
- added imports for "load_yaml_multi" and "load_yaml_dir" to
zenpacklib.py
- added "self.colorindex" to GraphPointSpec and "self.shorthand" to
RRDDatapointSpec in __init__, since missing these was causing Exceptions
with multiple YAML load/dump cycles (part of the validation testing)
- updated to accommodate logging changes

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
@@ -6,6 +6,7 @@ import time
 from .ZenPackLibLog import LOG
 from .Dumper import Dumper
 from .Loader import Loader
+import inspect
 
 
 # list of yaml sections with DEFAULTS capability
@@ -21,15 +22,30 @@ YAML_PREFERRED_ORDER = ['zProperties', 'class_relationships', 'classes',
                         'relationships']
 
 
+def get_calling_dir():
+    '''determine source directory of ZenPack's load_yaml call'''
+    frame = inspect.stack()[1]
+    for i in inspect.stack():
+        frame, filename, lineno, method_name, lines, idx = i
+        if '__init__.py' not in filename:
+            continue
+        for line in lines:
+            if 'load_yaml' in line:
+                return os.path.dirname(filename)
+    return None
+
+
 def load_yaml(yaml_doc=None, verbose=False, level=0):
     ''''''
     Loader.QUIET= not verbose
     Loader.LEVEL = level
 
+    # determine caller directory and attempt to load from it
     if not yaml_doc:
-        LOG.error("No YAML file specified")
-        return
-
+        try:
+            return load_yaml(get_calling_dir(), verbose, level)
+        except Exception as e:
+            LOG.error("YAML load error %s" % e)
     # loading from multiple files
     if isinstance(yaml_doc, list):
         # build python dict of merged YAML data
--- a/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
@@ -24,7 +24,6 @@ YAML_PREFERRED_ORDER = ['zProperties', 'class_relationships', 'classes',
 
 def get_calling_dir():
     '''determine source directory of ZenPack's load_yaml call'''
-    frame = inspect.stack()[1]
     for i in inspect.stack():
         frame, filename, lineno, method_name, lines, idx = i
         if '__init__.py' not in filename:
--- a/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
@@ -1,0 +1,256 @@
+import os
+import logging
+from collections import OrderedDict
+import yaml
+import time
+from .ZenPackLibLog import LOG
+from .Dumper import Dumper
+from .Loader import Loader
+
+
+# list of yaml sections with DEFAULTS capability
+YAML_HAS_DEFAULTS = ['classes', 'properties', 'thresholds', 'datasources',
+                     'datapoints', 'graphs', 'relationships','graphpoints']
+
+# preferred section ordering for YAML sections
+YAML_PREFERRED_ORDER = ['zProperties', 'class_relationships', 'classes',
+                        'device_classes', 'DEFAULTS', 'base',
+                        'label', 'order', 'properties',
+                        'monitoring_templates', 'dynamicview_views',
+                        'dynamicview_relations', 'impacts', 'impacted_by',
+                        'relationships']
+
+
+def load_yaml(yaml_doc=None, verbose=False, level=0):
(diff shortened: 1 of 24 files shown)


ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
```diff
@@ -0,0 +1,248 @@
...
+
+def sort_yaml_data(data):
+    """
+        Sort YAML data before outputting
+    """
+    if not isinstance(data, dict):
+        return data
+    ordered = OrderedDict()
+    fields = ['zProperties', 'class_relationships', 'classes', 'device_classes', 'DEFAULTS',
+              'base','label','order', 'properties', 'monitoring_templates',
```
dbouchillon: pep8 - needs a space between base, label, and order

Let's also make this a module global list so it only needs to be created once.  Less work for python, memory saved

ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    if not isinstance(data, dict):
+        return data
+    ordered = OrderedDict()
+    fields = ['zProperties', 'class_relationships', 'classes', 'device_classes', 'DEFAULTS',
+              'base','label','order', 'properties', 'monitoring_templates', 
+              'dynamicview_views', 'dynamicview_relations',
+              'impacts', 'impacted_by', 
+              'relationships']
+    if 'name' in data.keys():
```
dbouchillon: this could be moved within the for loop below.  also, you're calling data.keys() 3 times.  You can make a variable out of it, `data_keys = data.keys()`

ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def set_defaults(input):
+    """
+        if dict has multiple entries, look for identical key/value 
+        pairs and move them to DEFAULTS
+    """
+    defaults = {}
+    # get list of potential defaults
+    for name, params in input.items():
```
dbouchillon: should probably test that input.items() will always return name, params.  try/except should work

ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
```diff
@@ -0,0 +1,248 @@
...
+        if not isinstance(params, dict): 
+            continue
+        for k,v in params.items():
+            if isinstance(v, dict): 
+                continue
+            if k not in defaults.keys():
+               defaults[k] = v
+    # disqualify defaults if not universal
+    for k, v in defaults.items():
+        params = input.get(k)
```
dbouchillon: is this intentional?  params gets wiped out with the next statement

j053ph4: yeah that is OK b/c it's a separate for loop -- "inputs" is the only one that shouldn't be overwritten, and "defaults" gets modified throughout

ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
j053ph4: the "set_defaults" never gets called unless the "input" param is a dict, so it shouldn't be a problem here

ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
```diff
@@ -0,0 +1,257 @@
...
+        for name, params in input.items():
+            if not isinstance(params, dict):
+                continue
+            value = params.get(k)
+            if v != value:
```
dbouchillon: you could do this on one line

`if v != value and k in defaults.keys():`

ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
dbouchillon: what i mean is it's unnecessary as you're getting the params on the next line.

ZenPacks/zenoss/ZenPackLib/lib/spec/GraphPointSpec.py
```diff
@@ -82,6 +82,7 @@ def __init__(
         # Allow color to be specified by color_index instead of directly. This is
         # useful when you want to keep the normal progression of colors, but need
         # to add some DONTDRAW graphpoints for calculations.
+        self.colorindex = colorindex
```
dbouchillon: i don't see where this is used

j053ph4: i added this because repeated load/dump/load/dump cycles were failing when these didn't exist (there's another one too for RRDDatapointSpec 'shorthand')

ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
```diff
@@ -0,0 +1,256 @@
...
+                        'dynamicview_relations', 'impacts', 'impacted_by',
+                        'relationships']
+
+
+def load_yaml(yaml_doc=None, verbose=False, level=0):
+    ''''''
+    Loader.QUIET= not verbose
+    Loader.LEVEL = level
+
+    if not yaml_doc:
```
dbouchillon: The old version would default to using zenpack.yaml.  shouldn't we keep this ability?

j053ph4: That won't work unless there's some way to determine the directory of the __init__.py file calling the method (without passing __file__).  Do you know a way?

PR #227, opened 2016-08-19
Fixes error messages for components properties

ssvsergeyev:
Assume device components allowed to have properties named as in zenoss base classes

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -4951,7 +4951,8 @@ if YAML_INSTALLED:
         elif key in ZENOSS_KEYWORDS.union(JS_WORDS):
             # should be ok to use a zenoss word to define these
             # some items, like sysUpTime are pretty common datapoints
-            if cls not in [RRDDatasourceSpec,
+            if cls not in [ClassPropertySpec,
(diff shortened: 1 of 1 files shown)


j053ph4:
I think this warning is just to let the developer know that they are overwriting an attribute -- it doesn't stop it from being overwritten.  The assumption here is that those properties should be inherited, but nothing says it should always be so.  

Are you just wanting to suppress the warning?  Also, what ZenPack are you seeing this with?

ssvsergeyev:
@j053ph4 there are some related issues with these warnings: it shows every time ZP loaded, and in case of zenoss status it brings error as zenoss script use output of ZPs to get installed daemons (https://jira.zenoss.com/browse/ZEN-24417). We may escalate this to platform guys. Really I have no any objections about warnings :)

j053ph4:
OK I see that JIRA -- 2 more things to note:

1) in the "ZenPack" version of ZPL these types of messages will be disabled by default, and can be enabled on a per-ZP basis (develop mode)
2) Also, these messages are logged rather than being printed to screen as they are in the current dev build, and that will hopefully address the errors in the JIRA

PR #232, opened 2016-08-24
Fix logging and namespace issues preventing dependent ZenPack installation

j053ph4:
- ZenPackSpec allocates LOG class parameter to ZenPack class (removed
setting in __init__ (fixes installation issue)
- GraphPointSpec 'log' graph property (logarithm) was being overridden
by "log" logger instance, which prevents ZP installation from completing
correctly
- ZenPack.py now logs to stderr by default (during installation) with
verbosity settings in load_yaml()
- added stderr logging for TestCase and ZPLCommand classes
- changed 'log' argument to 'zplog' throughout
- classes should revert to default log if none specified
- removed unneeded imports
- fix string formatting in ZenPack.py
- move import of DEFAULTLOG from functions.py to helpers.ZenPackLog.py

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
@@ -21,7 +21,7 @@ from zope.browser.interfaces import IBrowserView
 from Products.ZenUI3.browser.interfaces import IMainSnippetManager
 from Products.ZenUI3.utils.javascript import JavaScriptSnippet
 from ..utils import dynamicview_installed
-from ..functions import get_symbol_name, get_zenpack_path, LOG
+from ..functions import get_symbol_name, get_zenpack_path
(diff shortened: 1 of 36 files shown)


ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
```diff
@@ -575,11 +572,13 @@ def create_zenpack_class(self):
             catalog = ".".join([self.name, class_]).replace(".", "_")
             attributes['GLOBAL_CATALOGS'].append('{}Search'.format(catalog))
 
-        return self.create_class(get_symbol_name(self.name),
+        cls = self.create_class(get_symbol_name(self.name),
                             get_symbol_name(self.name, 'schema'),
                             'ZenPack',
                             (ZenPack,),
                             attributes)
+        cls.LOG = self.LOG
```
j053ph4: here it is

PR #238, opened 2016-08-26
Ensure catalog creation respects spec property indexes (ZEN-18629)

j053ph4:
- Fixes ZEN-18269
- Revise get_catalog_scopes() to ignore the 'id' index if catalog is
other than "BaseComponentSearch" and if other indexes are defined


j053ph4:
Please review

PR #241, opened 2016-08-30
Ensure relation display falls back to inherited or relname_from_classname (ZEN-24302)

j053ph4:
- Auxillary fix for ZEN-24302
- addresses corner cases for extra_paths
- in cases where no relspec is found:
	- probe inherited_relationships()
	- fallback to relname_from_classname method


j053ph4:
Bump please -- this one needed for vSphere extra paths

j053ph4:
thanks!

PR #242, opened 2016-08-31
Disable reserved keyword errors for 1.1

j053ph4:
- will be moved to logging for 2.0
- done to avoid causing undue user concern in the short term

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -4997,8 +4997,9 @@ if YAML_INSTALLED:
     def yaml_warning(loader, e):
         # Given a MarkedYAMLError exception, either log or raise
         # the error, depending on the 'fatal' argument.
-
-        print format_message(e)
+        pass
+        # commenting out for 1.1 release
+        # print format_message(e)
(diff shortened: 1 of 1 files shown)


zenpacklib.py
```diff
@@ -4997,8 +4997,9 @@ def format_message(e):
     def yaml_warning(loader, e):
         # Given a MarkedYAMLError exception, either log or raise
         # the error, depending on the 'fatal' argument.
-
-        print format_message(e)
+        pass
+        # commenting out for 1.1 release
+        # print format_message(e)
```
pcarinhas: Will we still be able to get the warning during development? 
Is there some develop flag that we can set?
Perhaps log.debug() ?

j053ph4: Thanks for merging ... yes there are options like that, but since this is an "internal-only" release for shipping with OpenStack and VSphere I thought this the most expedient thing to do.

ZPL 2.0 will have more fine-grained event logging as well.

PR #260, opened 2016-10-06
Ensure catalog creation respects spec property indexes (ZEN-18269)

j053ph4:
- Fixes ZEN-18269
- Revise get_catalog_scopes() to ignore the 'id' index if catalog is
other than "BaseComponentSearch" and if other indexes are defined

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -2618,6 +2618,11 @@ class ClassSpec(Spec):
             if hasattr(base, '_catalogs'):
                 catalogs.update(base._catalogs)
 
+        if self.name not in catalogs:
+            catalogs[self.name] = {'indexes': {}}
+
+        indexes = catalogs[self.name]['indexes']
+
         # Add local properties and catalog indexes.
         for name, spec in self.properties.iteritems():
             if spec.api_backendtype == 'property':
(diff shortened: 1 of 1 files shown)


zenpacklib.py
```diff
@@ -2661,6 +2661,26 @@ def datapoint_method(self, default=spec.datapoint_default, cached=spec.datapoint
                     }
                 catalogs[self.name]['indexes'].update(pindexes)
 
+        def global_only():
```
cluther: Rather than adding all new code to fix the bad state catalogs has gotten into, I'd rather just replace the existing logic that's update catalogs to update it correctly to begin with. I think this would be a lot easier to follow when reading the code in the future, and this method is probably already too long to understand quickly.

Perhaps replacing the current _if pindexes:_ branch with this instead.

```python
# Add catalogs and indexes.
if pindexes:
    if self.name not in catalogs:
        catalogs[self.name] = {'indexes': {}}

    catalogs[self.name]['indexes'].update(pindexes)

    # Add a properly-scoped "id" index if it doesn't already exist.
    if 'id' not in catalogs[self.name]['indexes']:
        scopes = {
            x.get('scope', 'device')
            for x in catalogs[self.name]['indexes'].values()}

        catalogs[self.name]['indexes']['id'] = {
            'type': 'field',
            'scope': "device" if "device" in scopes else "global"}
```

zenpacklib.py
```diff
@@ -2651,15 +2656,13 @@ def datapoint_method(self, default=spec.datapoint_default, cached=spec.datapoint
...
-                        'indexes': {
-                            'id': {'type': 'field'},
-                        }
-                    }
-                catalogs[self.name]['indexes'].update(pindexes)
+            indexes.update(spec.catalog_indexes)
+
+        # Add a properly-scoped "id" index if needed
+        if len(indexes) > 0:
+            scopes = {j.get('scope', 'device') for i, j in indexes.items()}
```
cluther: _i_ is unused so you should use _indexes.values()_ instead.

```python
scopes = {x.get('scope', 'device') for x in indexes.values()}
```

zenpacklib.py
```diff
@@ -2651,15 +2656,13 @@ def datapoint_method(self, default=spec.datapoint_default, cached=spec.datapoint
...
+            indexes['id'] = {'type': 'field',
+                             'scope': 'device' if "device" in scopes else "global"}
```
cluther: Nitpick: Try to be consistent about single vs. double quoting.

PR #266, opened 2016-10-12
Support setting datapoint alias as string (ZEN-19486)

j053ph4:
- Fixes ZEN-19486
- when aliases is set as a string in YAML, add a dictionary to the
RRDDatapointSpec for the entry.
- added test_zen_19486 unit test


j053ph4:
bump please

PR #271, opened 2016-10-17
Ensure editable class property type setting in GUI (ZEN-22057)

j053ph4:
- Fixes ZEN-22057
- Suspect a platform issue, but unable to find it
- override __setattr__ method on ComponentBase to check value type
against _properties entry and set it
- An additional benefit will be guarding against invalid typed data from
modeler, scripts, dmd, etc.
- added test_zen_22057 unit test

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
@@ -49,6 +49,38 @@ class ComponentBase(ModelBase):
             }
         }
 
+    def __setattr__(self, name, value):
+        '''enforce type checking when setting _properties attributes'''
+        def lines(val):
+            """handle 'lines' type"""
+            if not isinstance(val, list):
+                val = [val]
+            val = [str(v) for v in val]
+            return val
+        # dictionary of _property "type" keys and target Python classes
+        target_type_map = {'boolean': bool,
+                           'int': int,
+                           'float': float,
+                           'string': str,
+                           'password': str,
+                           'lines': lines,
+                           }
+        # dictionary of _properties "id"s and "type"s
+        property_dict = {p.get('id'): p.get('type') for p in self._properties}
+        # only change type if it's a _properties attribute
+        if name in property_dict:
+            # and if we know how to handle it
+            target_class = target_type_map.get(property_dict.get(name))
+            if target_class:
+                try:
+                    value = target_class(value)
+                except Exception as e:
+                    self.LOG.warning('Error setting {} ({}) to: {} failed ({}).'.format(name,
+                                                                                        target_class.__name__,
+                                                                                        value,
+                                                                                        e))
+        super(ModelBase, self).__setattr__(name, value)
+
     def device(self):
         """Return device under which this component/device is contained."""
         obj = self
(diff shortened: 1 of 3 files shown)


ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
```diff
@@ -49,6 +49,32 @@ class ComponentBase(ModelBase):
             }
         }
 
+    def __setattr__(self, name, value):
+        '''enforce type checking when setting _properties attributes'''
+        # dictionary of _property "type" keys and target Python classes
+        target_type_map = {'boolean': bool,
```
cluther: The _lines_ type appears to be missing here. Is that intentional? I realize it would be slightly harder to handle than the other simple types, but verifying that it's a list of strings could be useful.

ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
```diff
@@ -49,6 +49,32 @@ class ComponentBase(ModelBase):
...
+                           'int': int,
+                           'float': float,
+                           'string': str,
+                           'password': str,
+                           }
+        # dictionary of _properties "id"s and "type"s
+        property_dict = dict([(p.get('id'), p.get('type')) for p in self._properties])
```
cluther: This would be better expressed as a dict comprehension.

```python
{p.get('id'): p.get('type') for p in self._properties}
```

ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
```diff
@@ -49,6 +49,32 @@ class ComponentBase(ModelBase):
...
+        # only change type if it's a _properties attribute
+        if name in property_dict.keys():
```
cluther: This nested checking of dictionary contents could be replaced with the following.

```python
target_class = target_type_map.get(property_dict.get(name))
if target_class:
    try:
        value = target_class(value)
    # etc..
```

In cases where a dict value can't be None it's usually better to "get" before "if" instead of "if in" before "get".

Also, there's no need to use the _keys_ method of a dict when doing an "in" check. It creates an unnecessary temporary list of the dictionaries keys when you can just do "if key in mydict" directly.

Lastly on the "except Exception", don't add the "as e" part if you aren't going to use e since it becomes an unused variable.

ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
```diff
@@ -281,7 +307,7 @@ def get_facets(self, root=None, streams=None, seen=None, depth=0, path=None, rec
                 # If 'all' mode, just include indirectly-related objects as well, in
                 # an unfiltered manner.
                 if recurse_all:
-                    for facet in obj.get_facets(root=root, seen=seen, path=path, depth=depth+1, recurse_all=True):
+                    for facet in obj.get_facets(root=root, seen=seen, path=path, depth=depth + 1, recurse_all=True):
```
cluther: No need to fix unrelated formatting in a functional change.

ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
```diff
@@ -293,7 +319,7 @@ def get_facets(self, root=None, streams=None, seen=None, depth=0, path=None, rec
                         if not recurse:
                             continue
 
-                        for facet in obj.get_facets(root=root, seen=seen, streams=[stream], path=path, depth=depth+1):
+                        for facet in obj.get_facets(root=root, seen=seen, streams=[stream], path=path, depth=depth + 1):
```
cluther: No need to fix unrelated formatting in a functional change.

PR #282, opened 2016-10-20
Fix ZenPack installation console logging

j053ph4:
- Fixes log messages not displaying to user during ZenPack
install/removal
- Added new logger instance with INFO logging to stderr
- fixed lack of space in log message


j053ph4:
bump please

PR #284, opened 2016-10-24
Only run certain unit tests if ZenPack dependencies are satisfied (ZEN-25690)

j053ph4:
- Fixes ZEN-25690
- Some unit tests depend on ZenPack-derived classes, so we cannot run
them if the dependencies are unsatisfied

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/tests/test_zen_24083.py
+++ b/ZenPacks/zenoss/ZenPackLib/tests/test_zen_24083.py
@@ -24,8 +24,17 @@ from Products.ZenTestCase.BaseTestCase import BaseTestCase
 from ZenPacks.zenoss.ZenPackLib.tests.ZPLTestHarness import ZPLTestHarness
 
 
+def aggregator_installed():
+    """Return True if AggregatingDataSource is installed"""
+    try:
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/ZenPackLib/tests/test_zen_24083.py
```diff
@@ -24,8 +24,17 @@
 from ZenPacks.zenoss.ZenPackLib.tests.ZPLTestHarness import ZPLTestHarness
 
 
+def aggregator_installed():
+    '''Return True if Duration Threshold is installed'''
```
dbouchillon: You should use double quotes for a docstring

ZenPacks/zenoss/ZenPackLib/tests/test_zen_25315.py
```diff
@@ -45,25 +45,35 @@
             parser: Auto
 """
 
+def win_shell_installed():
+    '''Return True if Duration Threshold is installed'''
```
dbouchillon: This doc string is confusing since you're loading Windows shell dataource.

Also, you should use double quotes here for the docstring

j053ph4:
OK I updated the docs

PR #297, opened 2016-10-27
Add proxies for all platform component classes

cluther:
ZenPacks have been created that wanted to extend platform component
classes such as IpInterface and FileSystem. The mechanism they used to
do this was to set the base classes to zenpacklib.Component first, and
something like Products.ZenModel.IpInterface.IpInterface second. This
resulted in buggy behavior because IpInterface behavior such as its
index_object() method would never be called.

Introducing zenpacklib proxies for all platform component classes solves
this problem because it puts a zenpacklib class as the single parent
class. This allows zenpacklib to properly fan out calls to things like
index_object() to all necessary parent classes.

Refs ZEN-25559.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@ Features
 * Improve performance of entity properties in component grids.
 * Simplify what device status means to critical event(s) in /Status.
 * Improve grid performance with streamlined info adapters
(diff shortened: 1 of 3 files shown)


dougsyer:
Awesomeness

PR #299, opened 2016-10-28
Use Spec/SpecParams init_params as a class property (10-25% reduced load time)

j053ph4:
- init_params is now a cached class property
- eliminates multiple runs of "init_params() method"
- corrected SpecParams overwrite of Spec init_params "type"

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/helpers/Dumper.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/helpers/Dumper.py
@@ -36,7 +36,7 @@ class Dumper(yaml.Dumper):
         be used for this specific zenpacklib.
     """
 
-    LOG=DEFAULTLOG
+    LOG = DEFAULTLOG
 
     def dict_representer(self, data):
         return yaml.MappingNode(u'tag:yaml.org,2002:map', data.items())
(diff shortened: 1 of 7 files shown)


ZenPacks/zenoss/ZenPackLib/lib/helpers/Dumper.py
```diff
@@ -67,7 +67,7 @@ def get_representation(self, value, v_type):
         elif v_type == 'Severity':
             return self.represent_str(self.severity_to_str(value))
         else:
-            
+
```
dbouchillon: if you're removing whitespace, let's just go ahead and remove the line

ZenPacks/zenoss/ZenPackLib/lib/params/SpecParams.py
```diff
@@ -8,36 +8,45 @@
...
             raise Exception("Spec Base Not Found for %s" % cls.__name__)
 
-        params = spec_base.init_params()
+        params = copy.deepcopy(spec_base.init_params)
         for p in params:
             params[p]['type'] = params[p]['type'].replace("Spec)", "SpecParams)")
 
-        return params
\ No newline at end of file
+        return params
```
dbouchillon: add a newline

PR #301, opened 2016-10-28
Fix failing Unit tests (ZEN-25766)

j053ph4:
- Fixes ZEN-25766
- fix exception in ClassSpec icon_url property, which was causing unit tests t ofail


j053ph4:
please review

PR #303, opened 2016-10-28
Input validation for GraphPoint color attribute

j053ph4:
Suggested methodology for performing input validation more generally
- Ensure that color is a 6-digit hexadecimal string
- Added Color class (string subclass) to perform validation
- Color class pads/truncates as needed, substituting 'F' for any non-hex
  character
- Added Dumper representer for Color
- added test_color_validation unit test


j053ph4:
This is a proposal for an approach for dealing with input validation generally.  I'd like to extend this to handle things like severity, modeler_lock, and other attributes that accept a narrow range of valid values.

PR #305, opened 2016-10-31
Update documentation for "extra_paths" parameter (ZEN-24084)

j053ph4:
- Fixes ZEN-24084
- Add notes and examples for using extra_paths


j053ph4:
doc update only -- please review

PR #306, opened 2016-10-31
Ensure unique id fields for subcomponent_nav_js (Fixes ZEN-25487)

j053ph4:
- Fixes ZEN-25487
- append meta_class to each "id" field in subcomponent_nav_js
- added test_zen_25487 unit test

Although I could not replicate the issue, I suspect that duplicate "id" parameters in the JS snippets could be to blame.  In any case, ensuring unique id fields should not cause any harm.


j053ph4:
please review -- minor change

j053ph4:
please reveiw

PR #311, opened 2016-11-04
Add proxies for all platform component classes

j053ph4:
ZenPacks have been created that wanted to extend platform component
classes such as IpInterface and FileSystem. The mechanism they used to
do this was to set the base classes to zenpacklib.Component first, and
something like Products.ZenModel.IpInterface.IpInterface second. This
resulted in buggy behavior because IpInterface behavior such as its
index_object() method would never be called.

Introducing zenpacklib proxies for all platform component classes solves
this problem because it puts a zenpacklib class as the single parent
class. This allows zenpacklib to properly fan out calls to things like
index_object() to all necessary parent classes.

Refs ZEN-25559.


ZenPacks/zenoss/ZenPackLib/lib/base/Component.py
```diff
@@ -6,9 +6,44 @@
...
+
 from ..factory.ComponentTypeFactory import ComponentTypeFactory
 
 
-Component = ComponentTypeFactory('Component', (DeviceComponent, ManagedEntity))
+Component = ComponentTypeFactory('Component', (BaseDeviceComponent, BaseManagedEntity))
+HWComponent = ComponentTypeFactory('HWComponent', (BaseHWComponent,))
+# Backwards-compatibility aliases.
+HardwareComponent = HWComponent
+CPU = ComponentTypeFactory('CPU', (BaseCPU,))
```
cluther: The "Backwards-compatibility aliases" comment above makes it appear that CPU and the following lines are also backwards-compatibility aliases. Perhaps it would be better to move these lines to the very bottom of the proxy class section to avoid that confusion.

```
# Backwards-compatibility aliases.
HardwareComponent = HWComponent
```

ZenPacks/zenoss/ZenPackLib/zenpacklib.py
```diff
@@ -42,8 +42,11 @@
 
 # these are the base classes for zenpacklib.Device, zenpacklib.Component, zenpacklib.HardwareComponent
 from lib.base.Device import Device
-from lib.base.Component import Component
-from lib.base.HardwareComponent import HardwareComponent
+from lib.base.Component import Component, HWComponent, HardwareComponent, CPU, \
```
cluther: I'd prefer to see the following multi-import style used in alphabetical order in cases like this.

```python
from lib.base.Component import (
    Component,
    HardwareComponent,
    HWComponent,
    etc,
    )
```

It's easier to visually scan for what's imported if they're sorted, and it creates much tidier diffs that result in fewer merge conflicts if they need to be changed.

PR #315, opened 2016-11-08
Revise class proxies Info and Interfaces classes (ZEN-25948)

j053ph4:
- added zuul.py with schema_map dictionary mapping each proxy class to
appropriate info and interface class
- updated ClassSpec create_info_schema_class and
create_iinfo_schema_class methods to use schema_map when determining
base classes
- fixed ImportError for HardwareComponent class (ZEN-25948)


j053ph4:
please review

PR #316, opened 2016-11-08
Fix errors in ZenPackSpec

j053ph4:
- remove unused import of memoize
- fix bad reference to create_zenpack_class method


j053ph4:
please review

PR #331, opened 2016-11-15
Support metric facade performance enhancements (ZEN-23870)

j053ph4:
* Fixes ZEN-23870
* Added "dataPointsToFetch" attribute to all generated Info classes
* ClassSpec datapoint_method uses "getFetchedDataPoint" instead of
"getRRDValue" when available
* Updated ClassSpec and RRDTemplateSpec to related associated datapoints
from "monitoring_templates"


j053ph4:
bump please

PR #332, opened 2016-11-16
Add support for specifying a device link provider

dbouchillon:
Fixes ZEN-26005

Supply provider using Module.Class format.  Getting the actual
links is very ZenPack specific, so there's not much to make generic
except for registering the adapter.


joshw:
I am a little concerned about encouraging the use of device link providers that can do anything.  I'd prefer a more limited type of functionality that could only search a catalog, and provide a ZPL way for zenpack developers to put their stuff into that catalog...

executing arbitrary code from a device link provider can potentially make a real mess out of system performance on all device overview pages, not just ones related to the zenpack...

PR #334, opened 2016-11-17
Support Impact Trigger specification (ZEN-26004)

j053ph4:
Impact Triggers


j053ph4:
OK this one is ready for review

ZenPacks/zenoss/ZenPackLib/lib/base/BaseTriggers.py
```diff
@@ -0,0 +1,52 @@
+from zope.interface import implements
```
dbouchillon: Let's go ahead and add the copyright header here

j053ph4: OK I added the copyright header

PR #336, opened 2016-11-21
Include zenpacklib TestCase for backwards compatibility

dbouchillon:
Fixes ZEN-26183

Redefine enableTesting.  This will only be called from a unit test
in another zenpack so it shouldn't cause zope clients to be in
test mode.

No need to try and instantiate a None.  could result in bad results

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/zenpacklib.py
+++ b/ZenPacks/zenoss/ZenPackLib/zenpacklib.py
@@ -33,6 +33,7 @@ __version__ = "2.0.0dev"
 # BaseTestCase which puts Zope into testing mode.
 TestCase = None
 
+import logging
 import Globals
 from Products.ZenUtils.Utils import unused
 unused(Globals)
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/ZenPackLib/zenpacklib.py
```diff
@@ -60,14 +61,108 @@
...
 from lib.spec import ZenPackSpec
 from lib.functions import relationships_from_yuml, catalog_search, ucfirst, relname_from_classname
 
+LOG = logging.getLogger('zen.ZenPackLib')
+
+
 def enableTesting():
-    """deprecated"""
-    pass
+    """This is for backwards compatibility.  For future tests, use ZPLTestHarness
```
joshw: I don't think this is really for backwards compatibility, as I understand it- enableTesting/TestCase is for regular unit tests in zenpacks.    ZPLTestHarness is for testing ZPL, right?

PR #337, opened 2016-11-28
Inform user about strictly enforced property types

dbouchillon:
Fixes ZEN-26183

We should alert the user that zpl2 enforces type.


dbouchillon:
bump

PR #339, opened 2016-11-30
Remove duplicate entries in subcomponent_nav_js (ZEN-26190)

j053ph4:
- Fixes ZEN-26190
- Added "remote_spec" attribute to ClassRelationshipSpec.
This provides a pointer so that the ClassRelationshipSpec knows its
remote class.  The "remote_class_spec" property method returns
remote_spec, unless its "class_" attribute is a descendant of
remote_class (inhertied relation), in which case the class_ spec is
returned.
- Revised the containing_components, containing_spec_relations methods
to use "containing_component_specs" method with parameter
- Rewrote containing_component_specs to probe spec relationships instead
of class _relationships
- Revised the faceting_components, faceting_spec_relations methods to
use "faceting_component_specs" method with parameter
- Rewrote faceting_component_specs to probe spec relationships instead
of class _relationships
- Revised subcomponent_nav_js method for clarity with comments.  Also
revised to ignore descendant classes to prevent duplicates.
- Updated test_subcomponent_nav_js unit test


j053ph4:
Updated to ensure dynamic view entries are not removed

j053ph4:
bump please

j053ph4:
pretty please review?

PR #341, opened 2016-12-02
Avoid overwriting modeler plugins in existing device class (ZEN-24901)

j053ph4:
- Fixes ZEN-24901
- Append rather than overwrite modeler plugins if device class already
exists
- Ensure old plugins remain when ZP removed
- Added "overwrite_plugins" to DeviceClassSpec to force overwrite
- Updated CHANGES and docs
- Added test_device_class_plugins unit test

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,7 @@ Features
 * Support for threshold graphpoint legend and color (ZEN-24904)
 * Ability to specify an initial sort column on a component grid
 * Performance enhancments for grid display of metrics (ZEN-23870)
+* Existing device class modeler plugins appended by default (ZEN-24901) 
 
 Fixes
 
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
@@ -285,9 +285,11 @@ class ZenPack(ZenPackBase):
         return diff
 
     def create_device_class(self, app, dcspec):
-        ''''''
+        """Return existing or created device class"""
+        exists = False
         try:
             dcObject = app.dmd.Devices.getOrganizer(dcspec.path)
+            exists = True
         except KeyError:
             self.LOG.info('Creating DeviceClass {}'.format(dcspec.path))
             dcObject = app.dmd.Devices.createOrganizer(dcspec.path)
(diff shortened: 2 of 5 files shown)


j053ph4:
bump please

CHANGES
```diff
@@ -27,6 +27,7 @@ Features
 * Support for threshold graphpoint legend and color (ZEN-24904)
 * Ability to specify an initial sort column on a component grid
 * Performance enhancments for grid display of metrics (ZEN-23870)
+* existing device class modeler plugins appended by default (ZEN-24901)
```
cluther: Use consistent case in CHANGES.

ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -285,16 +285,33 @@ def template_changed(self, app, existing, new_spec):
         return diff
 
     def create_device_class(self, app, dcspec):
-        ''''''
+        """Return existing or created device class"""
+
+        def handle_plugins(d_class, value, overwrite):
```
cluther: handle_plugins shouldn't be an inner function. It's only called once within the function.

ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -285,16 +285,33 @@ def template_changed(self, app, existing, new_spec):
...
             dcObject = app.dmd.Devices.getOrganizer(dcspec.path)
+            exists = True
         except KeyError:
             self.LOG.info('Creating DeviceClass {}'.format(dcspec.path))
             dcObject = app.dmd.Devices.createOrganizer(dcspec.path)
 
         for zprop, value in dcspec.zProperties.iteritems():
             if dcObject.getPropertyType(zprop) is None:
                 self.LOG.error("Unable to set zProperty {} on {} (undefined zProperty)".format(zprop, dcspec.path))
+            # modeler plugins are a special case for preexisting device class
```
cluther: Why are modeler plugins a special case? Aren't there are zProperties that we wouldn't want to overwrite?

j053ph4: I thought it was possible that this could also apply to any "list" or "lines" type of zProperty, but I thought I should not do anything here since there hasn't been any specific demand and I wanted to avoid unanticipated consequences.

j053ph4: If you think this will come up in other contexts, then perhaps I should make this parameter an attribute of ZPropertySpec instead of DeviceClassSpec?

PR #345, opened 2016-12-05
Avoid setting zProperties on existing device classes

j053ph4:
- Addresses (ZPS-137)
- zProperties will no longer be set on preexisting device classes
- Updated CHANGES and yaml-device-classes.rst regarding these changes

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,19 @@
 Version 2.0
 ===========
 
+Backwards Incompatible Changes
+
+* Any installed ZenPacks using older versions of zenpacklib.py will continue to function unchanged.
+* Using version 2.0 is slightly different.  The __init__.py file import statements should now contain the following:
+
+.. code-block:: python
+
+    from ZenPacks.zenoss.ZenPackLib import zenpacklib
+    CFG = zenpacklib.load_yaml()
+    schema = CFG.zenpack_module.schema
+
+* zProperties will not be updated automatically on existing device classes.  These should be handled on a case basis by using migrate scripts.
+
 Release 2.0.0
 -------------
 
@@ -29,6 +42,7 @@ Features
 * Performance enhancments for grid display of metrics (ZEN-23870)
 * Support for Device Link Providers
 * Added troubleshooting aid for easily saving function data(writeDataToFile)
+* Avoid setting zProperties on existing device class (ZPS-137)
 
 Fixes
 
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
@@ -290,13 +290,20 @@ class ZenPack(ZenPackBase):
 
     def create_device_class(self, app, dcspec):
         ''''''
+        exists = False
         try:
             dcObject = app.dmd.Devices.getOrganizer(dcspec.path)
+            exists = True
         except KeyError:
             self.LOG.info('Creating DeviceClass {}'.format(dcspec.path))
             dcObject = app.dmd.Devices.createOrganizer(dcspec.path)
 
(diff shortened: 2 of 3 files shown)


j053ph4:
fixed conflicts please review

CHANGES
```diff
@@ -1,5 +1,19 @@
 Version 2.0
 ===========
+Backwards Incompatible Changes
```
cluther: Preserve the existing convention of an empty line between section title and its content.

CHANGES
```diff
@@ -1,5 +1,19 @@
...
+------------------------------
+* Any installed ZenPacks using older versions of zenpacklib.py will continue to function unchanged.
```
cluther: _Backwards Incompatible Changes_ can probably just be made regular text like _Features_ and _Fixes under each release as lists immediately after section titles don't look very good. You'd normally want a regular paragraph following a section title.

There needs to be a blank line before beginning a list in reStructuredText.

CHANGES
```diff
@@ -1,5 +1,19 @@
...
+* Using version 2.0 is slightly different.  The __init__.py file import statements should now contain the following:
+
+.. code-block:: python
+
+	from ZenPacks.zenoss.ZenPackLib import zenpacklib
```
cluther: Follow the existing convention of 4 spaces of indentation for code-block content.

CHANGES
```diff
@@ -1,5 +1,19 @@
...
+	CFG = zenpacklib.load_yaml()
+	schema = CFG.zenpack_module.schema
```
cluther: Why is it that .schema is no longer valid?

CHANGES
```diff
@@ -1,5 +1,19 @@
...
+
+* zProperties will not be updated automatically on existing device classes.  These should be handled on a case basis by using migrate scripts
```
cluther: List items should either consistently end with periods or not.

CHANGES
```diff
@@ -1,5 +1,19 @@
...
+* Further documentation of these and other changes can be found at `<http://zenpacklib.zenoss.com/en/latest/>`_.
```
cluther: I'm not sure that stating that you can read more about the changes by reading all of the documentation is useful.

CHANGES
```diff
@@ -29,6 +43,7 @@ Features
 * Performance enhancments for grid display of metrics (ZEN-23870)
 * Support for Device Link Providers
 * Added troubleshooting aid for easily saving function data(writeDataToFile)
+* Avoid setting zProperties on existing device class(ZPS-137)
```
cluther: Missing space before ending parenthetical.

ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -290,13 +290,19 @@ def template_changed(self, app, existing, new_spec):
...
             dcObject = app.dmd.Devices.getOrganizer(dcspec.path)
+            exists = True
         except KeyError:
             self.LOG.info('Creating DeviceClass {}'.format(dcspec.path))
             dcObject = app.dmd.Devices.createOrganizer(dcspec.path)
 
         for zprop, value in dcspec.zProperties.iteritems():
+            # Avoid setting zProperties on an existing device class
+            if exists:
+                self.LOG.debug('Not setting "{}" to "{}" on existing device class ({})'.format(zprop, value, dcspec.path))
```
cluther: This log should probably only be produced if the YAML has a different value than the device class.

Also, you may want to consider using the `!r` formatted rather than creating your own quotes.

```python
"Not setting {} to {!r} on {}".format(zprop, value, dcspec.path)
```

CHANGES
j053ph4: there's nothing that exports it to the namespace i guess -- want to take a look?

PR #348, opened 2016-12-07
Support for user-added components (REVIEW ONLY)

j053ph4:
- ClassPropertySpec now uses "editable" to find definable fields and
"optional" to filter them
- Added facade_class, ifacade_class, and router_class to ClassSpec
- Added ClassSpec register_user_defined method
- Added ClassSpec add_component_js, js_add_fields, get_editable to
provide JS code for add menu
- Updated Spec get_class_factory to handle IFacade
- Updated ZenPackSpec to create new JS and to register router classes
- Updated documentation to cover new parameters
- Added "test_user_added_component" to unit tests.
- Updated CHANGES

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
@@ -23,7 +23,7 @@ from Products.ZenModel.interfaces import IExpandedLinkProvider
 from Products.ZenUI3.utils.javascript import JavaScriptSnippet
 from ..utils import dynamicview_installed
 from ..functions import get_symbol_name, get_zenpack_path
-from ..resources.templates import JS_LINK_FROM_GRID
+from ..resources.templates import JS_LINK_FROM_GRID, CONFIG_ZCML
 from ..gsm import get_gsm
 from ..base.Device import Device
 from ..base.ZenPack import ZenPack
@@ -153,6 +153,7 @@ class ZenPackSpec(Spec):
         self.plumb_properties()
 
         # Class Relationship Schema
+        self.imported_classes = {}
         self.class_relationships = []
         if class_relationships:
             if not isinstance(class_relationships, list):
@@ -243,6 +244,8 @@ class ZenPackSpec(Spec):
         self.create_ordered_component_tree()
         self.create_global_js_snippet()
         self.create_device_js_snippet()
+        self.create_add_js_snippet()
+        self.register_user_resources()
         self.register_browser_resources()
         self.apply_platform_patches()
         self.register_link_providers()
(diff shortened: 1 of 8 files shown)


docs/yaml-classes-and-relationships.rst
```diff
@@ -621,6 +621,12 @@ extra_paths
 .. todo:: Add section on Impact & DynamicView.
 
 .. todo:: Add more detailed explanation of extra_paths, based on comments in zenpacklib.py
+  
+user_defined
```
cluther: I can't intuit what a class being user_defined would mean, and I think there might be better options for this property. Though the description below is great.

Some alternatives to user_defined:
- user_addable
- allow_user_creation
- add_components_menu

ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
```diff
@@ -466,6 +482,25 @@ def create_device_js_snippet(self):
         return self.create_js_snippet(
             'device', snippet, classes=device_classes)
 
+    def create_add_js_snippet(self):
+        """Register component add JavaScript snippet."""
+        device_classes = [
```
cluther: Treating all device types within the same ZenPack to the same components on their add menu will probably end up getting us in trouble. Such as in the Windows ZenPack where we have Device and ClusterDevice, or someday in cases like CiscoUCS where we have Device and CIMCDevice.

ZenPacks/zenoss/ZenPackLib/lib/functions.py
```diff
@@ -24,6 +24,18 @@
 
 # Private Functions #########################################################
 
+def object_from_string(name, text):
```
cluther: I can't find object_from_string or add_class_text_method being used, nor can I find any of the new string templates being used. Am I missing something, or is this not completely implemented yet?

I'm not sure I like the idea of creating code this way. It makes debugging even harder, and it makes editors unable to help with code issues.

ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
j053ph4: How about an additional check that the relation exists on the device classes?

ZenPacks/zenoss/ZenPackLib/lib/functions.py
j053ph4: I forgot to remove add_class method, and will move object_from_string to Spec

docs/yaml-classes-and-relationships.rst
```diff
@@ -702,6 +708,12 @@ editable
   :Type: boolean
   :Default Value: false
   
+optional
+  :Description: If component is user_edited, and editable is true, is this an optional attribute?
```
j053ph4: The idea here was that the component might have many (editable) properties, only some of which the developer would want to be in the dialog box.  The "optional" would give a potential override by narrowing the fields that would be on the form

cluther: If I understand what you're saying, you wanted this "optional" boolean to control whether a field shows up on the add dialog or not. It has nothing to do with editing on the details panel?

If so, I'd name it "add_dialog_display" to continue the theme of "grid_display" and "details_display".

ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
cluther: Yes, that makes sense. Though it raises another point about only components that have a direct contained-by relationship to the device to be added in this way. It's worth thinking about what happens if someone enables "user_defined" on a component class that doesn't meet this criteria, or if they enable "user_defined" on a Device subclass.

j053ph4: Can you take a look at the check I put in there and see if it makes sense?

Also, I restricted the "user_defined" stuff to components only, and the menu items only get added if the to-one is a device (since the js is per-device class).

Meaning this feature isn't available for and won't work for a component 1:MC addable_component, only for a device 1:MC addable_component.

PR #351, opened 2016-12-08
Fix to help with circular dependencies

dbouchillon:
Fixes ZPS-319

LinuxMonitor and OpenStackInfrastructure have circular dependencies
causing each zenpack to load the other.  This will keep a dict of
loaded CFGs so that we only need to load in a zenpack's yaml once.

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/zenpacklib.py
+++ b/ZenPacks/zenoss/ZenPackLib/zenpacklib.py
@@ -39,7 +39,8 @@ from Products.ZenUtils.Utils import unused
 unused(Globals)
 
 # ZenPacks use these to load their YAML files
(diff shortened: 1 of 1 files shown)


cluther:
After reviewing ZPS-319 I think there are three problem, and none of them are addressed by this pull request.

1. There was a problem handling DEFAULTS in the spec. Handled in #350.
2. There's a problem in OSI's feature/zpl2 branch where patches/platform.py is attempting to import `..zenpacklib` which isn't a thing anymore.
3. There's a circular import between OSI's __init__.py and LinuxMonitor's __init__.py.

2 and 3 need to be fixed. Neither needs to be fixed in zenpacklib.

PR #352, opened 2016-12-08
Streamline and better document "order" parameter (ZPS-124)

j053ph4:
The "order" parameter is now expected to be an integer between 1 and 100
for classes, properties, and relationships with a default of 50.  This
was the documented behavior but the implementation was less consistent.
 
Float arguments are still accepted, but normalized to integer between
1-100.

- Added methods to spec to deal with setting, normalizing, and scaling
the order parameter where appropriate.
- Implemented float detection to determine if order parameter on
properties, relationships, or classes needs to be renormalized.
- Added a note explaining order parameter behavior to documentation.

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/Spec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/Spec.py
@@ -9,6 +9,8 @@
 import inspect
 import re
 import logging
+import itertools
+import operator
 from collections import OrderedDict
 
 from Products import Zuul
@@ -143,6 +145,7 @@ class Spec(object):
     source_location = None
     speclog = None
     _init_params = None
+    _order = None
     LOG = DEFAULTLOG
 
     def __init__(self, _source_location=None, zplog=None):
@@ -170,6 +173,34 @@ class Spec(object):
             parts.append(super(Spec, self).__str__())
 
         return "{}({})".format(self.__class__.__name__, ' - '.join(parts))
+
+    @property
+    def order(self):
+        """order attribute used by ClassSpec, ClassRelationSpec, ClassPropertySpec"""
+        return self._order
+
+    @order.setter
+    def order(self, value):
+        """order should be an integer between 1 and 100"""
+        # deal with legacy cases
+        if isinstance(value, int):
+            self._order = int(max(0, min(100, value)))
+        else:
+            self._order = value
+
+    def normalize_child_order(self, items):
+        """Deal with legacy order given as float 
+        or with default value (100)
+        """
+        # case where any order is a float or all are default=100
+        if any(isinstance(x.order, float) for x in items) or all(x.order == 100 for x in items):
+            new_orders = itertools.chain(xrange(1, 101), itertools.repeat(100))
+            for c in sorted(items, key=operator.attrgetter("order")):
+                c.order = int(new_orders.next())
+
+    def scale_order(self, scale=1, offset=5):
+        """Return linear order transform"""
+        return scale * float(self.order) / 100 + offset
 
     def apply_data_defaults(self, dictionary, default_defaults=None, leave_defaults=False):
         """Modify dictionary to put values from DEFAULTS key into other keys.
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
@@ -146,6 +146,8 @@ class ZenPackSpec(Spec):
 
         # Classes
         self.classes = self.specs_from_param(ClassSpec, 'classes', classes, zplog=self.LOG)
+        # deal with float order parameters if they exist
+        self.normalize_child_order(self.classes.values())
 
         # update properties from ancestor classes
         self.plumb_properties()
@@ -289,7 +291,7 @@ class ZenPackSpec(Spec):
             if x.is_device}
 
         order = {
-            x.meta_type: float(x.order)
+            x.meta_type: x.scaled_order
             for x in self.classes.itervalues()}
 
         def getComponentTree(self, uid=None, id=None, **kwargs):
--- a/docs/yaml-classes-and-relationships.rst
+++ b/docs/yaml-classes-and-relationships.rst
@@ -500,11 +500,32 @@ initial_sort_column
   :Default Value: name
 
 order
-  :Description: Order to display this class among other classes. (0-100)
+  :Description: Order to display this class among other classes. (1-100)
   :Required: No
   :Type: integer
-  :Default Value: 50
+  :Default Value: 100
   
+.. note::
+
+      The *order* parameter takes any integer value between 1 and 100.
+      However, it's behavior depends somewhat depending on whether it applies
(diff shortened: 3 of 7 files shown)


ZenPacks/zenoss/ZenPackLib/lib/spec/Spec.py
```diff
@@ -171,6 +172,55 @@ def __str__(self):
...
+            if i.normalize_order:
+                return True
+        return False
+
+    def normalize_child_order(self, items):
+        """Deal with legacy order parameters given as float"""
+        if self.normalize_children(items):
+            candidates = [p for p in items if p.order != 50]
+            values = [c.order for c in candidates]
+            factor = .5
```
cluther: This normalization doesn't work very well for at least the CiscoUCS ZenPack. Many different floats that are intentionally larger and smaller than each other get normalized to the same integer values.

For example:
- 4.1 -> 45
- 4.2 -> 45
- 4.3 -> 45
- 4.4 -> 45
- 4.5 -> 57
- 4.5 -> 57

There's a typical approach to changing the scale of a set of numbers that would fare a little better, but it's still subject to duplicate results for very-close inputs.

```python
def rescale(values, new_min=1, new_max=100):
    if len(values) < 2:
        return map(int, values)
    old_min = min(values)
    old_max = max(values)
    a = (new_max - new_min) / (old_max - old_min)
    b = new_max - a * old_max
    return [a * x + b for x in values]
```

A more fool-proof approach for this particular case is probably to just sort then sequence.

```python
new_orders = itertools.chain(xrange(1, 101), itertools.repeat(100))
for c in sorted(candidates, key=operator.attrgetter("order"):
    c.order = new_orders.next()
```

ZenPacks/zenoss/ZenPackLib/lib/spec/ZenPackSpec.py
```diff
@@ -289,7 +291,7 @@ def create_ordered_component_tree(self):
             if x.is_device}
 
         order = {
-            x.meta_type: float(x.order)
+            x.meta_type: float(x.scaled_order)
```
cluther: Wrapping this in float is no longer required. x.scaled_order is guaranteed to be a float.

ZenPacks/zenoss/ZenPackLib/lib/spec/Spec.py
```diff
@@ -171,6 +174,43 @@ def __str__(self):
...
+        """Return True if order should be renormalized"""
+        return isinstance(self.order, float)
+
+    def normalize_children(self, items):
+        """Return True if order should be normalized"""
+        return any(x.normalize_order for x in items)
+
+    def normalize_child_order(self, items):
+        """Deal with legacy order parameters given as float"""
+        if self.normalize_children(items):
```
cluther: Given that normalize_children is a single-line method, and this is the only place it's used, you may want to inline the its code then eliminate the normalize_children method.

ZenPacks/zenoss/ZenPackLib/lib/spec/Spec.py
```diff
@@ -171,6 +174,43 @@ def __str__(self):
...
+            candidates = [p for p in items if p.order != 50]
+            midline = 50 - int(round(len(candidates) / 2))
```
cluther: I don't understand what midline's purpose. It would seem to result in something like this..

- 1st prop in list of 10 props: 46
- 10th prop in list of 10 props: 55
- 1st prop in list of 200 props: -49
- 200th prop in list of 200 props: 150

docs/yaml-classes-and-relationships.rst
```diff
@@ -500,11 +500,32 @@ initial_sort_column
...
+      this order:
+      
+      1. Containing Components
+      2. Properties
+      3. Contained Components
+
+      with container relationships listed before visible properties and finally any
+      containing relationships. 
+      
+      Earlier (pre-2.0) versionsof ZenPackLib accepted float arguments for order.  However,
```
cluther: Missing space in _versionsof_.

docs/yaml-classes-and-relationships.rst
```diff
@@ -569,7 +590,7 @@ dynamicview_weight
   :Description: Dynamic View weight for objects of this class. Higher numbers are further to the right. Can be overridden by implementing getDynamicViewGroup() method on class.
   :Required: No
   :Type: float or int
-  :Default: 1000 + (order * 100)
+  :Default: 1000 * (1 + order/100)
```
cluther: This formula is confusing to me. Though it looks to match what the code does. Perhaps it would be a bit easier to understand if stated as `1000 + (order * 10)`.

ZenPacks/zenoss/ZenPackLib/lib/spec/Spec.py
```diff
@@ -171,6 +174,36 @@ def __str__(self):
...
+    def order(self, value):
+        """order should be an integer between 1 and 100"""
+        # deal with legacy cases
+        if isinstance(value, int):
+            self._order = int(max(0, min(100, value)))
+        else:
+            self._order = value
+
+    @property
+    def normalize_order(self):
```
cluther: It looks like normalize_order isn't used anymore.

docs/yaml-classes-and-relationships.rst
```diff
@@ -569,7 +590,7 @@ dynamicview_weight
   :Description: Dynamic View weight for objects of this class. Higher numbers are further to the right. Can be overridden by implementing getDynamicViewGroup() method on class.
   :Required: No
   :Type: float or int
-  :Default: 1000 + (order * 100)
+  :Default: 1000  + order * 10
```
cluther: extra space

docs/yaml-classes-and-relationships.rst
```diff
@@ -816,10 +837,10 @@ grid_display
   :Default Value: true
   
 order
-  :Description: Order to display this relationship among other relationships and properties. (0-10)
+  :Description: Order to display this relationship among other relationships and properties. (1-100)
   :Required: No
-  :Type: float
-  :Default Value: 3.0 for To-One, 6.0 for To-Many.
+  :Type: integer
+  :Default Value: 50
```
cluther: default value appears to now be 100

docs/yaml-classes-and-relationships.rst
```diff
@@ -691,10 +712,10 @@ grid_display
   :Default Value: true
   
 order
-  :Description: Order to display this property among other properties. (0-100)
+  :Description: Order to display this property among other properties. (1-100)
   :Required: No
   :Type: integer
-  :Default Value: 45
+  :Default Value: 50
```
cluther: default value appears to now be 100

docs/yaml-classes-and-relationships.rst
```diff
@@ -500,11 +500,32 @@ initial_sort_column
   :Default Value: name
 
 order
-  :Description: Order to display this class among other classes. (0-100)
+  :Description: Order to display this class among other classes. (1-100)
   :Required: No
   :Type: integer
   :Default Value: 50
```
cluther: default value appears to now be 100

PR #356, opened 2016-12-12
Ensure correct device status icon for up/down

j053ph4:
Fixes ZPS-78

It was assumed that the zep event filter used in the getStatus method
override (DeviceBase) was recursive for /Status and its sub-event
classes.  Since it is not, a list of event classes is included in the
query event filter.


j053ph4:
please review.

PR #357, opened 2016-12-12
Updates for event class spec in zpl

dbouchillon:
The 'create' flag is essentially redundant.  if a user is
specifying an event class then they obviously want it to be created.
For the removal we just need to know if it was created by the zpl.
Also we are making sure not to remove platform event classes.

Fixes ZPS-433


dbouchillon:
bump

PR #362, opened 2016-12-14
Fix ZenPack install template change detection inadequacies

j053ph4:
- Found while reviewing ZPS-102
- Template upgrade changes not being detected when upgrading between
.egg files
- Moved template detection to "install" method from "remove" method
since the "remove" method uses the old (existing) specs for comparison
(thus finding no difference).
- replaced the "template_changed" method with more general
"object_changed" method that can handle other types of objects
- added "object_changed_safe" method which, although less accurate,
removes the intermediate dummy object creation/removal (so it may be
better suited for device classes and other organizer specs).
- modifed RRDTemplateSpec "create" method to accept optional object id
parameter
- added RRDTemplateSpec "remove" method to faciliate object removal

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/tests/test_template_modify.py
+++ b/ZenPacks/zenoss/ZenPackLib/tests/test_template_modify.py
@@ -23,7 +23,6 @@ from Products.ZenTestCase.BaseTestCase import BaseTestCase
 
 # zenpacklib Imports
 from ZenPacks.zenoss.ZenPackLib.tests.ZPLTestHarness import ZPLTestHarness
-from ZenPacks.zenoss.ZenPackLib.lib.base.ZenPack import ZenPack
 
 
 YAML_DOC = """name: ZenPacks.zenoss.ZenPackLib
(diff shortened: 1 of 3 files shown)


ZenPacks/zenoss/ZenPackLib/tests/test_template_modify.py
```diff
@@ -303,14 +302,20 @@ def get_result(self, orig_doc, new_doc, expected=None):
...
+
+        zenpack = z_new.schema.ZenPack(z_new.dmd)
         # new temlate spec
         new_tspec = z_new.cfg.device_classes.get('/Server').templates.get('Device')
-        diff = zenpack.template_changed(z_new, orig_template, new_tspec)
+        new_tspec_param = zenpack._v_specparams.device_classes.get('/Server').templates.get('Device')
+
+        diff = zenpack.object_changed(z_new.dmd, orig_template, new_tspec, new_tspec_param)
+
+        # diff = zenpack.template_changed(z_new, orig_template, new_tspec)
```
dbouchillon: you should go ahead and remove this line

PR #364, opened 2016-12-14
Update to class override logging and documentation (ZPS-112)

j053ph4:
- Fixes ZPS-112
- ClassSpec updated to ensure that per-ZP logging extends to created
classes (and used by class overrides).
- Documentation updated with a note about logging verbosity.


j053ph4:
bump please

PR #365, opened 2016-12-15
Update example __init__ used in zenpacklib --create

j053ph4:
- removed os import and filename/path specification (no longer needed)
- added optional "verbose" and "level" parameters for demonstration 
- added "schema" line to facilitate class overrides.
- Updated README.txt to reflect latest usage


j053ph4:
bump please -- minor update

PR #373, opened 2016-12-19
zenpacklib --create should add some conventional subdirectories

j053ph4:
- updated "create" method to include building out some typically used
subdirectories such as migrate, tests, modeler/plugins, etc...

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/libexec/ZPLCommand.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/libexec/ZPLCommand.py
@@ -284,6 +284,18 @@ class ZPLCommand(ZenScriptBase):
                 namespace_init_f.write(
                     "__import__('pkg_resources').declare_namespace(__name__)\n")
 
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/ZenPackLib/lib/libexec/ZPLCommand.py
```diff
@@ -284,6 +284,18 @@ def create_zenpack_srcdir(self, zenpack_name):
...
+                   'migrate', 'resources', 'modeler', 'tests',
+                   'libexec', 'modeler/plugins', 'lib']
+        for subdir in subdirs:
+            dirpath = os.path.join(module_directory, subdir)
+            os.makedirs(dirpath)
+            init_fname = os.path.join(dirpath, '__init__.py')
+            print "  - creating file: {}".format(init_fname)
+            with open(init_fname, 'w') as init_f:
+                init_f.write("\n")
+
```
dbouchillon: can you add a `init_f.close()` here?

PR #387, opened 2017-01-11
Fix inaccurate datapoint YAML dump (ZEN-26593)

j053ph4:
- Fixes ZEN-26593
- Corrects issue where datapoint aliases inadvertently removed from YAML
output
- Updated Dumper to use "use_shorthand" method in addition to existence
test
of 'shorthand' when outputting to YAML


j053ph4:
I updated CHANGES regarding this issue

PR #389, opened 2017-01-12
Fix potential traceback during ZP install

j053ph4:
- Added device class retrieval and logging if device class not found
- Error could occur if DeviceClassSpec "create" attribute is set to
False, but target device class doesn't exist.


cluther:
Next time bump the version on the hotfix branch immediately after creating it in its own commit.

The version shouldn't be changed along with functional changes. It can make things uglier down the road such as cherry-picking or reverting commits.

PR #390, opened 2017-01-12
Fix ZP installation reattempts (ZPS-627)

j053ph4:
- Fixes ZPS-627
- Corrects issue where if ZenPack installation terminates prematurely,
backup files are left behind

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -24,6 +24,7 @@ Release 2.0.1
 Fixes
 
 * Ensure all datapoint attributes export to YAML (ZEN-26593)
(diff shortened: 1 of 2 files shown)


CHANGES
```diff
@@ -24,6 +24,7 @@ Release 2.0.1
 Fixes
 
 * Ensure all datapoint attributes export to YAML (ZEN-26593)
+* Ensure subsquent installations complete if ZP install fails (ZPL-627)
```
cluther: Should be ZPS-627 instead of ZPL-627.

j053ph4: Good catch -- fixed.

PR #397, opened 2017-01-17
Ensure proxy class instances displayed in GUI (ZPS-651) REVIEW ONLY

j053ph4:
- Fixes ZPS-651 (and related ZPS-572)
- override facade _instanceClass property method so that it returns a
tuple of ZPL-derived classes in addition to the ZenModel base class.
- This alters the facade methods so that they return the expected
instances (components) in the various GUI elements
- Example is OSProcess-subclass instances in the Processes tab


j053ph4:
bump please

PR #398, opened 2017-01-17
Avoid repeated calls to get_symbol_name

j053ph4:
- Small optimization since this function only needs to be called twice.


j053ph4:
bump please

PR #399, opened 2017-01-19
Reorganization and Refactoring of Unit Tests

j053ph4:
- Created ZPLTestBase, ZPLTestComand, LogCapture base classes for use by
ZPLTestHarness and related tests
- Changed subclasses of most tests to ZPLTestBase or ZPLTestCommand
- Removed unused imports and redundant code
- Removed some redundant tests
- Removed dependency on "real" zope instance for all but 4 tests:
	- test_datasource_defaults_inheritance.py
	- test_impact_triggers.py
	- test_installed_zenpacks.py
	- test_usepowershell_datasource.py
- Removed superfluous logging on all but "test_install" tests
- Replaced "TestClasses" with "TestInstalledZenPacks" which, instead of
validating against included YAML files (which is problematic in
different scenarios), performs validation against all ZenPackLib
2.0-based ZenPacks installed on the test-running system.


j053ph4:
please review

PR #406, opened 2017-02-02
Add missing "-" to zenpack create command

stmcginnis:
The command given to create a new ZenPack was missing the "--" before "create", causing the create to fail and print out the help message.


cluther:
@stmcginnis: Thanks for the report. The zenpacklib command line argument change from "create" to "--create" was done going from zenpacklib v1 to v2, and the HTTP API tutorial wasn't updated accordingly.

PR #412, opened 2017-02-15
Allow custom threshold graph point id (ZPS-855)

j053ph4:
- Allow GraphPointSpec threshold graph point to have id other than
threshold id
- Allows faithful specification threshold graph points in YAML

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/GraphPointSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/GraphPointSpec.py
@@ -9,6 +9,7 @@
 from Products.ZenModel.GraphPoint import GraphPoint
 from Products.ZenModel.DataPointGraphPoint import DataPointGraphPoint
 from Products.ZenModel.ComplexGraphPoint import ComplexGraphPoint
+from Products.ZenModel.ThresholdGraphPoint import ThresholdGraphPoint
 from ..base.types import Color
 from .Spec import Spec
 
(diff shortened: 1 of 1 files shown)


j053ph4:
please review

ZenPacks/zenoss/ZenPackLib/lib/spec/GraphPointSpec.py
```diff
@@ -133,9 +134,14 @@ def thresholdLegends(self, value):
...
             if not isinstance(data, dict):
-                data = {'legend': None, 'color': None}
+                data = {'legend': None, 'color': None, 'threshId': None}
             self._thresholdLegends[id]['legend'] = data.get('legend')
             self._thresholdLegends[id]['color'] = data.get('color')
+            self._thresholdLegends[id]['threshId'] = data.get('threshId', id)
+
+    def get_threshold_ids(self):
+        """find thresholds related to this datapoint"""
+        return [t.name for t in self.template_spec.template_spec.thresholds.values() if self.dpName in t.dsnames]
```
pcarinhas: Doesn't one need to check existence for elements in the chained expression: 
self.template_spec.template_spec.thresholds.values ?

j053ph4: no it's not needed those will always exist -- the template_specs are required for this class' existence, and the 'thresholds' will always be an OrderedDict even if empty. -- just checked to be sure

PR #417, opened 2017-02-20
Add a simple ZP name check

vikt0rs:
Added a check, that ZP name starts with "ZenPacks." to avoid typos and
errors like https://github.com/zenoss/ZenPacks.zenoss.ZenPackLib/issues/415

### DIFF
--- a/zenpacklib.py
+++ b/zenpacklib.py
@@ -6315,6 +6315,13 @@ def create_zenpack_srcdir(zenpack_name):
     import shutil
     import errno
 
+    # validate ZP name - it must begins with "ZenPacks.*"
+    if not zenpack_name.startswith('ZenPacks.'):
+        sys.exit(
+            "`{}` is a wrong name for ZenPack - zenpack name must starts "
+            "with `ZenPacks`. Please, check ZP name.".format(zenpack_name)
+        )
+
     if os.path.exists(zenpack_name):
         sys.exit("{} directory already exists.".format(zenpack_name))
 


zenpacklib.py
```diff
@@ -6315,6 +6315,13 @@ def create_zenpack_srcdir(zenpack_name):
     import shutil
     import errno
 
+    # validate ZP name - it must begins with "ZenPacks.*"
+    if not zenpack_name.startswith('ZenPacks.'):
+        sys.exit(
+            "`{}` is a wrong name for ZenPack - zenpack name must starts "
+            "with `ZenPacks`. Please, check ZP name.".format(zenpack_name)
+        )
```
ssOleg: I would suggest changing message to something like this:
"ZenPack name is incorrect, it does not match ZenPack naming requirenments. Please see documentation: https://zenpack-sdk.zenoss.com/en/stable/managing-zenpacks-4.html#naming-a-zenpack"

vikt0rs: I think it's better to explain the error in place - not provide some external link to some some "magic" `ZenPack naming requirements`.

j053ph4: I agree that it's a good idea to perform this check and also to provide direct feedback.  The main issue with this PR, though, is that the monolithic zenpacklib.py file is now vestigial and is not the appropriate place for the change.  

I took your suggestion and created a similar change here:
https://github.com/zenoss/ZenPacks.zenoss.ZenPackLib/pull/420
and also added a check for the length of the ZenPack namespace (since that might go wrong as well).

vikt0rs: ok, I'll close this PR in favour of #420

PR #420, opened 2017-02-22
Perform name check to avoid ZenPack creation errors

j053ph4:
When creating a ZenPack, the user-supplied name should be checked for
compliance with ZenPack naming conventions.  Added a method do perform a
check and to provide the user with guidance regarding the convention.

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/libexec/ZPLCommand.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/libexec/ZPLCommand.py
@@ -227,9 +227,24 @@ class ZPLCommand(ZenScriptBase):
         except Exception, e:
             DEFAULTLOG.exception(e)
 
+    def validate_zenpack_name(self, zenpack_name):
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/ZenPackLib/lib/libexec/ZPLCommand.py
```diff
@@ -227,10 +227,25 @@ def lint(cls, filename):
         except Exception, e:
             DEFAULTLOG.exception(e)
 
+    def validate_zenpack_name(self, zenpack_name):
+        """Ensure that ZenPack name conforms with convention"""
+        zenpack_name_parts = zenpack_name.split('.')
```
vikt0rs: If we are working with ZP name parts below, we can use something similar to the following construction:

```
try:
    zp_name_1, zp_name_2, zp_name_3 = zenpack_name.split('.')
except ValueError:
    sys.exit('ZenPack name must conform ... ')
```

PR #422, opened 2017-02-23
Preserve order when loading from multiple YAML files (DO NOT MERGE)

j053ph4:
- added BasicLoader class to handle OrdereDict and sequenced data
- subclassed ZPL Loader from BasicLoader
- load_yaml_single now uses BasicLoader if "useLoader" is False
- added "sorted" boolean argument to "get_optimized_yaml"
- restricted "sorting" behavior to "optimize_yaml" method (used by
--optimized flag)
- changed ZenPack "get_yaml_diff" to classmethod for convenience
- updated test_load_dir to print out YAML diff when failing


cluther:
Closing. See #423 instead.

PR #425, opened 2017-02-24
Update unit tests to display YAML diff if they fail

j053ph4:


cluther:
This pull request is conflicting with hotfix/2.0.3 now. Is any of it still needed?

PR #434, opened 2017-03-02
Fix missing Dynamic View on some components

cluther:
Some ZenPacks such as Docker have custom component types, but no custom
device type. They do this by patching their component's relationship
onto ZenModel.Device. The problem that can then occur is multiple
component Dynamic View (subcomponent_view) Component navigations can get
created for a single device type. The two views will disagree about
which types of components should have a Dynamic View. The user will
always end up seeing this as components not having a DynamicView in
their Display drop-down when they should.

This fix namespaces the subcomponent_view, so it's OK to have multiple
registered for the same device.

Refs ZPS-703.

The actual fix for ZPS-703 is in the Docker ZenPack. This commit is to
incorporate the fix into ZenPackLib proper.


joshw:
approve

PR #435, opened 2017-03-09
Fix for not creating device classes in some cases

cluther:
The current develop branch of the Windows ZenPack is experiencing a
problem with ZenPack where the /Server/Microsoft/Windows device class
doesn't get installed when the ZenPack is installed.

This issue seems to have been brought to light by the fix for ZEN-925 in
ZenPackLib 2.0.3. Specifically 07cf5e5.

Acquisition strikes again. When the following device classes exist..

- /Server/Windows
- /Server/Microsoft

The following code will return the /Server/Windows device class.

    dmd.Devices.getOrganizer("Server/Microsoft/Windows")

Obviously this isn't what we want, and it results in ZenPackLib thinking
that it doesn't need to create the /Server/Microsoft/Windows device
class because it already exists.

This fix should address that problem by adding an additional check to
after the call to getOrganizer() that makes sure the returned device
class has the same path as what was asked for.

There's a bit of reorganization in this change to put all of the
organizer specs under a new abstract base class called OrganizerSpec so
that the ability to safely get an organizer can be shared.

Fixes ZPS-1012.

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
@@ -71,7 +71,7 @@ class ZenPack(ZenPackBase):
         # load monitoring templates
         for dcname, dcspec in self.device_classes.iteritems():
             dcspecparam = self._v_specparams.device_classes.get(dcname)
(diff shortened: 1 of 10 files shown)


dbouchillon:
i'm dave and i approve this pull request

PR #437, opened 2017-03-10
Fix event mappings with id/eventClassKey mismatch

cluther:
Event class mappings with an eventClassKey different than their id were
getting indexed while their id and eventClassKey were still identical
prior to zenpacklib setting the custom eventClassKey. This resulted in
the correct eventClassKey not being indexed in
dmd.Events.eventClassSearch, and therefore the mappings not matching any
events.

This simple fix indexes mappings after all of their properties have been
set to solve the problem.

Fixes ZPS-1016. Fixes part of ZPS-900.

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/EventClassMappingSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/EventClassMappingSpec.py
@@ -73,4 +73,5 @@ class EventClassMappingSpec(Spec):
             if getattr(mapping, x) != getattr(self, x):
                 setattr(mapping, x, getattr(self, x, None))
 
-        self.zpl_managed = True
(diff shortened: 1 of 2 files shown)


joshw:
The fact that this is a hotfix makes me think that there may be zenpacks which triggered this problem in the wild, and which would thus have things missing from the catalog.     If that's right, I think we should include a migration script to attempt to fix the damage.  The apparent mistake in the zpl_managed flag, if that was present in previously released versions, may make that difficult to do in an efficient manner, though.

cluther:
@joshw: I'm thinking it's not that big of a concern. We (Zenoss) haven't shipped any ZenPacks yet that use `event_classes` in their YAML.

joshw:
You may merge if you think that's the case.

PR #442, opened 2017-03-21
Ensure that template changes are detected during pre-2.0 upgrades

j053ph4:
- Fixes ZPS-787
- Updated ZenPack install and remove methods to account for both ZPL
2.0-based and legacy zenpack-provided templates
- Change detection now works across versions
- New methods added to make the associated logic more contained and
consistent (and hopefully more easily understood)
- Methods should be abstract enough to later apply to other ZPL-provided
objects like event classes, process sets, etc.
- Changed various 'dmd' references to app.zport.dmd for readability

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
@@ -47,9 +47,9 @@ class ZenPack(ZenPackBase):
     def __init__(self, *args, **kwargs):
         super(ZenPack, self).__init__(*args, **kwargs)
 
-    def _buildDeviceRelations(self, batch=10):
+    def _buildDeviceRelations(self, app, batch=10):
         count = 0
-        for d in self.dmd.Devices.getSubDevicesGen():
+        for d in app.zport.dmd.Devices.getSubDevicesGen():
             d.buildRelations()
             count += 1
             if count % batch == 0:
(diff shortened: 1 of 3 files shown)


ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -190,25 +149,129 @@ def remove(self, app, leaveObjects=False):
...
+                if backup_target:
+                    self.move_object(parent, relname, backup_target.id, object_id)
+                # or in the case of the existing object, leave it alone
+                else:
+                    pass
+
+    def check_diff(self, app, parent, relname, object, spec, specparam):
+        """Return True if object has changed creating preupgrade backup if needed"""
+        diff = self.object_changed(app, object, spec, specparam)
+        if diff:
```
cluther: You could rearrange this a bit to eliminate a level of indentation for the method.

```python
if not diff:
    return False

# preserve the existing object if different
# ..etc.
```

ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -190,25 +149,129 @@ def remove(self, app, leaveObjects=False):
...
+        """Attempt to retrieve an object given its id, parent instance, and relation name"""
+        rel = getattr(parent, relname, None)
+        if rel:
+            try:
+                return rel._getOb(object_id)
+            except AttributeError:
+                pass
+        return None
+
+    def get_or_create_backup(self, parent, relname, object_id, overwrite=True):
```
cluther: I don't see the overwrite argument actually being used. I'd stick to the [YAGN principle](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) here and not create code paths that aren't used.

ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -190,25 +149,129 @@ def remove(self, app, leaveObjects=False):
...
+            try:
+                rel.manage_renameObject(source_id, dest_id)
+            except Exception as e:
+                LOG.warn("Could not move template {}/{} to {}/{} ({})".format(parent.getDmdKey(), source_id,
+                                                                              parent.getDmdKey(), dest_id,
+                                                                              e))
+        else:
+            LOG.warn("Cannot rename {} on nonexistent relation {}".format(source_id, relname))
+
+    def move_object(self, parent, relname, source_id, dest_id, overwrite=True):
```
cluther: I don't see the overwrite argument actually being used. I'd stick to the [YAGN principle](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) here and not create code paths that aren't used.

ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
j053ph4: OK yeah left that in there for posterity -- will remove

PR #448, opened 2017-03-31
Revise testing for extra_params handling

j053ph4:
- Replaces several unit tests with generalized examples
- Eliminates dependency on external zenpacks for this testing
- Provides a common methodology for this sort of testing
- Eliminate remaining dmd connectivity from other unit tests
- Remove ZenScriptBase from ZPLTestHarness, move related functions to ZPLTestHarnessScriptBase


ZenPacks/zenoss/ZenPackLib/tests/test_zenpacklib_version.py
```diff
@@ -0,0 +1,50 @@
...
+from ZenPacks.zenoss.ZenPackLib.tests.ZPLTestBase import ZPLTestBase
+from ZenPacks.zenoss.ZenPackLib import zenpacklib
+
+
+class TestInstalledZenPackLibVersion(ZPLTestBase):
+    """Test this version of ZenPackLib against relevant installed ZenPacks"""
+
+    def get_version(self):
+        '''extract the ZenPack version from setup.py'''
+        path = os.path.join(os.path.dirname(__file__), '..', '..', '..', '..', 'setup.py')
```
cluther: This seems like it would fail in the product-assembly tests. There's no setup.py in that case because the ZenPack is installed as an egg, not in link mode.

j053ph4:
OK fixed that issue with the test_zenpacklib_version unit test

PR #450, opened 2017-04-06
Fix missing paths from getPaths() issue (ZPS-1262)

j053ph4:
- Fixes ZPS-1262

Two main issues are involved here.  First, the various PathReporters
used by ZPL proxy classes need to have their own getPaths() overrides
like the one ZPL provides.

Second, the ComponentBase containing_relname method fails for proxy
classes whose primary parents are OperatingSystem or Hardware, since
these are only valid if the subclass has the same name as the parent
class (like FileSystem).  Otherwise, the device is directly related to
the class.

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
@@ -231,6 +231,13 @@ class ComponentBase(ModelBase):
         """Return name of containing relationship."""
         for relname, relschema in self._relations:
             if issubclass(relschema.remoteType, ToManyCont):
+                # skip if it's not the real containing rel
+                # for instance an inherited os.filesystems for sub-classed FileSystem
+                rel = getattr(self, relname, None)
+                if rel:
+                    if not rel.obj:
+                        continue
+
                 return relname
         raise ZenSchemaError("%s (%s) has no containing relationship" % (self.__class__.__name__, self))
 
--- a/ZenPacks/zenoss/ZenPackLib/lib/wrapper/ComponentPathReporter.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/wrapper/ComponentPathReporter.py
@@ -9,23 +9,59 @@
 from zope.component import adapts
 from zope.interface import implements
 from Products.Zuul.catalog.interfaces import IPathReporter
-from Products.Zuul.catalog.paths import DefaultPathReporter, relPath
+from Products.Zuul.catalog.paths import (
+     DefaultPathReporter,
+     DevicePathReporter,
+     ServicePathReporter as ServicePathReporterBase,
+     InterfacePathReporter as InterfacePathReporterBase,
+     ProcessPathReporter as ProcessPathReporterBase,
+     ProductPathReporter as ProductPathReporterBase,
+     relPath
+     )
+
 from ..base.ComponentBase import ComponentBase
+from ..base.Component import (
+     HWComponent,
+     IpInterface,
+     OSProcess,
+     Service,
+)
 
 
-class ComponentPathReporter(DefaultPathReporter):
-
-    """Global catalog path reporter adapter factory for components."""
-
+class ComponentPathMixIn(DefaultPathReporter):
+    """Global catalog path reporter override"""
     implements(IPathReporter)
-    adapts(ComponentBase)
 
     def getPaths(self):
-        paths = super(ComponentPathReporter, self).getPaths()
+        paths = super(ComponentPathMixIn, self).getPaths()
 
         for facet in self.context.get_facets():
             rp = relPath(facet, facet.containing_relname)
             paths.extend(rp)
 
         return paths
+
+
+class ComponentPathReporter(ComponentPathMixIn):
+    """Global catalog path reporter adapter factory for basic components."""
+    adapts(ComponentBase)
+
+
+class ProductPathReporter(ComponentPathMixIn, ProductPathReporterBase):
+    """"""
+    adapts(HWComponent)
+
+
+class InterfacePathReporter(ComponentPathMixIn, InterfacePathReporterBase):
+    """"""
+    adapts(IpInterface)
+
+
+class ProcessPathReporter(ComponentPathMixIn, ProcessPathReporterBase):
+    """"""
+    adapts(OSProcess)
+
+class ServicePathReporter(ComponentPathMixIn, ServicePathReporterBase):
+    """"""
+    adapts(Service)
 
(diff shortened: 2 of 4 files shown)


ZenPacks/zenoss/ZenPackLib/lib/wrapper/ComponentPathReporter.py
```diff
@@ -9,23 +9,59 @@
...
 
         return paths
 
+
+class ComponentPathReporter(ComponentPathMixIn):
+    """Global catalog path reporter adapter factory for basic components."""
+    adapts(ComponentBase)
+
+
+class ProductPathReporter(ProductPathReporter, ComponentPathMixIn):
```
cluther: You shouldn't redefine ProductPathReporter like this. It changes the meaning of what ProductPathReporter is from the beginning to the end of the module. It would be better to import ProductPathReporter as ProductPathReporterBase to eliminate that potential confusion.

I don't think this adapter is going to work as expected. When ProductPathReporter.getPaths() is called, it will end up calling ProductPathReporterBase.getPaths(), not ComponentPathMixin.getPaths() because ProductPathReporterBase is the first superclass, and it has a getPaths method.

These same issues also exist for InterfacePathReported, ProcessPathReported, and ServicePathReporter below.

ZenPacks/zenoss/ZenPackLib/lib/zuul.py
```diff
@@ -59,12 +59,13 @@
 from Products.Zuul.facades.processfacade import ProcessFacade
 from Products.Zuul.facades.servicefacade import ServiceFacade
 
-from Products.Zuul.catalog.paths import (
-     DevicePathReporter,
+from Products.Zuul.catalog.paths import DevicePathReporter
```
cluther: This and the following change make no functional difference, and they're the only changes to this file. It would be better to leave them out of this commit. They only distract from the real changes.

j053ph4: Actually these changes are needed since the imports are coming from the ZPL overrides for each of the path reporters

j053ph4:
please review

j053ph4:
bump please

ZenPacks/zenoss/ZenPackLib/lib/base/ComponentBase.py
```diff
@@ -231,6 +231,13 @@ def get_containing_relname(self):
         """Return name of containing relationship."""
         for relname, relschema in self._relations:
             if issubclass(relschema.remoteType, ToManyCont):
+                # skip if it's not the real containing rel
+                # for instance an inherited os.filesystems for sub-classed FileSystem
+                rel = getattr(self, relname, None)
```
joshw: Can you explain this a bit more?  How is that not a "real" containing relationship?

j053ph4: This gets into the fact that ZPL by default roots component relations to the device, not the os or hw subcomponents of the device.  This created a problem for the ZPL Proxy subclasses of the traditional classes like FileSystem or OsComponent. The ZPL suclass roots these at device.filesystems(), whereas ZenModel puts them at device.os.filesystems().

This in effect means that the proxy components have multiple containing relations (ZPL and ZenModel), which resulted the need for the additional existence test for the rel.obj before returning the actual "real" containing relname (the one that will actually work)

joshw: You can't have multiple containing relationships on an object though. That would be totally invalid, would it not?

j053ph4: Yes that's correct in one sense (they are invalid/non-functioning) but are a side effect of using the proxy classes.  So far it has only surfaced here.  One potential alternative fix would be for ZPL to defer to ZenModel's containing relations, but I bet that would have all kinds of implications.

j053ph4: So far this is the quickest fix w/least impact AFAIK

joshw: I would be pretty concerned about having contradictory info in _relations.  I can not imagine that not biting us later.

j053ph4: Understood but at the moment this is the only manifestation of an issue and it's a minor one.  I am estimating a fair amount of time needed to understand the scope of impact, what the options are,  and whether any cure would be worse than the disease.

The issue was originally reported by Doug S. and took a few hours to unravel.

dbouchillon: what about having an overriding property in the relationship override fields to make this optional?

joshw: I think the only way to work around the issue and have a valid schema is to purge the other containing relationship from _relations when a new one is added by ModelTypeFactory.  At least then there is a schema that makes sense, and shouldn't break anything. But conceptually it's still weird- i don't know why we would have the ZPLified objects live in a non-default location like that.   Unless we can justify that, we really should be thinking about how to fix this so things live in a normal location.  Not an easy thing to fix now, but it will only get harder the longer we leave it like that.  It's possible that nobody but doug is using these proxies, and in that case we may still be able to put the genie back in the bottle..

I would say that ModelTypeFactory (and possibly ClassSpec as well) need to enforce a few things about relationships:
   - warning or error when changing the schema of an inherited relationship name - this is technically allowable, but almost certainly a bad idea.
   - error when creating a doubly-contained class - this is probably not technically allowed, and definitely a bad idea.

dougsyer:
Right now im only using the hwcomponent to get the relation to the productkeys.  I coukd just as easily usr a zplib component that just has that relaiom like what was there in 1.x.

Im not using proxies for anything else..

I could live without the proxies as they are today and truth be told they kinda scare me.

Im about to zenoss5ify everyting anyways.

Of course winzp is using them im assuming.

I do need to be able non containing rels to ipinterface and i should be using the os process classes for a couple projects..im sure i could deal with that using old spis

dougsyer:
I dont think we can update to stock zplib 2.06 or 2.07+ without this fix or guidance on how to handle the issue with hwcomponent  in zenpacklib not working as it did in zplib version 1.0.

dougsyer:
the ComponentBase code broke the deletion of windows services and processes we think.  we were able to finally reproduce it in our dev system.  we applied this patch to 2.07 but did not apply the component change and it seems like the modeling issue is fixed AND the windows services /processes problem is gone also...

PR #456, opened 2017-04-28
Fix ZEN-27142: Typecast empty str to int

alam-zenoss:


j053ph4:
I'd like to hold off approving this until we can duplicate the issue and make sure that this fixes it

dougsyer:
I think i ran into a zenpack install issue when someone added a ds/dp intona zenpacklib template but had rrdmin set to the default of None.  I had to set the rrdmins to string 0 to go forward with the install.  Not sure id this is a fix for that issue but.  To reproduce add a ds dp and then try to reinstall.  I havent tried that yet but it was tripping up on the integer check for rrdmin and the zenpack was set to 0 i think it may have been the manually created dp

PR #457, opened 2017-05-04
Use an appropriate sequence for all of the Graph Points.

itshane:


Hemie143:
Why not the enumerate function ? It's more Pythonic.

itshane:
The sequence must be incremental and unique within all of the graph points, including the comment graph points.

PR #459, opened 2017-05-11
Ensure that Severity of 0 (Clear) is respected (ZPS-1454)

j053ph4:
- Fixes ZPS-1454
- Update existence testing of values to include zero values
- Updated unit test to account for this scenario

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/types.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/types.py
@@ -112,7 +112,7 @@ class Severity(int):
                'clear': 0}
 
     def __new__(cls, value):
-        if not value:
+        if not value and not isinstance(value, int):
             return value
         return int.__new__(cls, cls.validate(value))
 
--- a/ZenPacks/zenoss/ZenPackLib/lib/helpers/Dumper.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/helpers/Dumper.py
@@ -200,13 +200,13 @@ class Dumper(yaml.Dumper):
     def represent_severity(self, data):
         """represent Severity"""
         orig = getattr(data, 'orig')
-        if orig:
+        if orig is None:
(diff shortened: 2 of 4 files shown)


j053ph4:
bump please

ZenPacks/zenoss/ZenPackLib/lib/base/types.py
```diff
@@ -112,7 +112,7 @@ class Severity(int):
                'clear': 0}
 
     def __new__(cls, value):
-        if not value:
+        if not value and not isinstance(value, int):
```
dbouchillon: seems like we should actually just allow for any value to flow through to validate() and then issue a warning there saying it's invalid?  if it's not a valid int or str, then we could default it to 3 and issue the warning.  or even raise an exception

```
    def __new__(cls, value):
        return int.__new__(cls, cls.validate(value))

    @classmethod
    def validate(cls, value):
        try:
            value = int(value)
            if value < 0:
                cls.LOG.warning("Invalid severity value ({}), increasing to 0".format(value))
                value = 0
            elif value > 5:
                cls.LOG.warning("Invalid severity value ({}), reducing to 5".format(value))
                value = 5
        except (TypeError, ValueError):
            if isinstance(value, str) and value.lower() in cls._valid_text:
                value = cls.to_num.get(value.lower())
            else:
                # return 3 or raise exception here?
                sev_num_txt = [str(x) for x in cls._valid_num]
                cls.LOG.warning("Invalid severity value ({}), "\
                    "must be one of: ({}) or ({}).".format(value,
                                            ', '.join(cls._valid_text),
                                            ', '.join(sev_num_txt)))
                value = 3
        return value
```

j053ph4: This won't work as described because int.__new__() gives an error:

    int() argument must be a string or a number, not 'NoneType'

if the "value" parameter is None (which is default if unset).

PR #462, opened 2017-06-06
Remove obsoleted unit tests from root (ZPS-1292)

j053ph4:
- Possible fix for ZPS-1292
- All but one converted for use with ZenPackLib 2
- These files are possibly involved with build failures


dbouchillon:
changes requested

GNUmakefile
```diff
@@ -1,6 +1,6 @@
 ##############################################################################
 #
-# Copyright (C) Zenoss, Inc. 2015, all rights reserved.
+# Copyright (C) Zenoss, Inc. 2017, all rights reserved.
```
dbouchillon: need to include 2015 also.  so 2015-1017

GNUmakefile
```diff
@@ -13,5 +13,8 @@ help:
 	@echo "Please use 'make <target>' where <target> is one of.."
 	@echo "  test    Runs all tests in tests/ directory."
 
+egg:
```
dbouchillon: ok, there actually needs to be more to egg building.  i think in setup.py we'd need to make some changes.  just remove this part for now.  keep the test section though

dbouchillon:
approved

j053ph4:
thanks

PR #465, opened 2017-06-21
Allow creation of thresholds to be optional

yichi-lu:
Fixes ZPS-1666

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
@@ -22,6 +22,7 @@ class RRDThresholdSpec(Spec):
             eventClass=None,
             severity=None,
             enabled=None,
+            optional=False,
             extra_params=None,
             _source_location=None,
             zplog=None
@@ -41,6 +42,8 @@ class RRDThresholdSpec(Spec):
             :param enabled: TODO
             :type enabled: bool
             :param extra_params: Additional parameters that may be used by subclasses of RRDDatasource
+            :param optional: is this threshold optional?
+            :type optional: bool
             :type extra_params: ExtraParams
 
         """
@@ -58,6 +61,7 @@ class RRDThresholdSpec(Spec):
         self.severity = Severity(severity)
         self.enabled = enabled
         self.type_ = type_
+        self.optional = optional
         if extra_params is None:
             self.extra_params = {}
         else:
@@ -74,9 +78,13 @@ class RRDThresholdSpec(Spec):
 
         threshold_types = dict((y, x) for x, y in template.getThresholdClasses())
         type_ = threshold_types.get(self.type_)
+
         if not type_:
-            raise ValueError("'%s' is an invalid threshold type. Valid types: %s" %
-                             (self.type_, ', '.join(threshold_types)))
+            if self.optional:
+                return
+            else:
+                raise ValueError("'%s' is an invalid threshold type. Valid types: %s" %
+                                 (self.type_, ', '.join(threshold_types)))
 
         threshold = template.manage_addRRDThreshold(self.name, self.type_)
         self.speclog.debug("adding threshold")
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,6 +4,16 @@
 Changes
 #######
 
+Version 2.1
+===========
+
+Release 2.1.0
+-------------
+
+Features
+
+* Add "optional" field for thresholds (ZPS-1666)
+
 Version 2.0
 ===========
 
--- a/docs/yaml-monitoring-templates.rst
+++ b/docs/yaml-monitoring-templates.rst
@@ -548,6 +548,12 @@ escalateCount:
   :Type: int
   :Default Value: None
 
+optional:
+  :Description: The threshold will not be created if the threshold type is not available and *optional* is set to true. Installation will fail if the type is not available and *optional* is set to false.
+  :Required: No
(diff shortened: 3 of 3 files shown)


ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
```diff
@@ -74,25 +74,29 @@ def create(self, templatespec, template):
...
+            if self.enabled is not None:
+                threshold.enabled = self.enabled
+            if self.extra_params:
+                for param, value in self.extra_params.iteritems():
+                    if param in [x['id'] for x in threshold._properties]:
+                        setattr(threshold, param, value)
+                    else:
+                        raise ValueError("%s is not a valid property for threshold of type %s" % (param, type_))
+        else:
+            if self.extra_params.get('optional', False):
```
cluther: _optional_ should be a first-class parameter for RRDThresholdSpec, not in extra_params. extra_params are for parameters that some threshold types might have, and others not.

ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
```diff
@@ -74,25 +74,29 @@ def create(self, templatespec, template):
...
+                threshold.severity = int(self.severity)
+            if self.enabled is not None:
+                threshold.enabled = self.enabled
+            if self.extra_params:
+                for param, value in self.extra_params.iteritems():
+                    if param in [x['id'] for x in threshold._properties]:
+                        setattr(threshold, param, value)
+                    else:
+                        raise ValueError("%s is not a valid property for threshold of type %s" % (param, type_))
+        else:
```
cluther: Style-wise, I'd rather see this else section moved up right under `type_ = threshold_types.get(self.type_)` as `if not type_:` and just return after logging if _optional_ was True. That way the normal-path logic doesn't have to be indented under an if.

ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
```diff
@@ -74,25 +74,29 @@ def create(self, templatespec, template):
...
+            if self.extra_params.get('optional', False):
+                self.LOG.info("%s thresholds were not created, since %s ZP was not installed" % (self.type_, self.type_))
```
cluther: I think we decided we don't want to log anything in the optional=True case when the threshold type is not available. This is not something the user should be aware of.

ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
```diff
@@ -24,6 +24,7 @@ def __init__(
             enabled=None,
             extra_params=None,
             _source_location=None,
+            optional=False,
```
cluther: _optional_ should come before extra_params.

ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
```diff
@@ -62,6 +65,7 @@ def __init__(
             self.extra_params = {}
         else:
             self.extra_params = extra_params
+        self.optional = optional
```
cluther: Should probably move it before extra_params here too. Just for consistency. Obviously it doesn't matter for functionality.

docs/changes.rst
```diff
@@ -4,6 +4,16 @@
...
 
+Version 2.1
+===========
+
+Release 2.1.0
+-------------
+
+Features
+
+* Add optional field for thresholds.
```
cluther: It might make sense to quote or otherwise highlight _optional_ here. Otherwise the obvious reading would be that some optional field was added, but who knows what it was. We want to convey that a field named "optional" was added.

docs/yaml-monitoring-templates.rst
```diff
@@ -548,6 +548,12 @@ escalateCount:
   :Type: int
   :Default Value: None
 
+optional:
+  :Description: indicating whether this threshold is optional or not
```
cluther: Description should be a proper sentence or sentences to match the convention used in the other descriptions. It should also impart additional information than the field name. I don't think this current description helps people understand what "optional" means and what the effect of setting it to true or false would be.

Maybe something more like..

> This threshold will be skipped during installation if the type of threshold isn't available on the system when _optional_ is set to true. Installation will fail if the type of threshold isn't available when _optional_ is set to false.

ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
```diff
@@ -22,8 +22,10 @@ def __init__(
             eventClass=None,
             severity=None,
             enabled=None,
+            optional=False,
             extra_params=None,
             _source_location=None,
+            optional=False,
```
cluther: You forgot to remove the second _optional_.

ZenPacks/zenoss/ZenPackLib/lib/spec/RRDThresholdSpec.py
```diff
@@ -58,10 +62,12 @@ def __init__(
         self.severity = Severity(severity)
         self.enabled = enabled
         self.type_ = type_
+        self.optional = optional
         if extra_params is None:
             self.extra_params = {}
         else:
             self.extra_params = extra_params
+        self.optional = optional
```
cluther: Forgot to remove this too.

PR #466, opened 2017-08-01
Added missed SpecParams imports, improved get_init_params method to r…

itshane:
…eturn appropriate init params.

Fixes ZPS-1601

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/params/EventClassSpecParams.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/params/EventClassSpecParams.py
@@ -1,6 +1,6 @@
 ##############################################################################
 #
-# Copyright (C) Zenoss, Inc. 2016, all rights reserved.
+# Copyright (C) Zenoss, Inc. 2016-2017, all rights reserved.
 #
 # This content is made available according to terms specified in
 # License.zenoss under the directory where your Zenoss product is installed.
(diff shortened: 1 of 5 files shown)


ZenPacks/zenoss/ZenPackLib/lib/params/EventClassSpecParams.py
```diff
@@ -1,6 +1,6 @@
 ##############################################################################
 #
-# Copyright (C) Zenoss, Inc. 2016, all rights reserved.
+# Copyright (C) Zenoss, Inc. 2017, all rights reserved.
```
cluther: Could you make the copyright 2016-2017 instead of removing the 2016? I think it's more important to include the original date than it is to include new dates.

itshane: Done.

ZenPacks/zenoss/ZenPackLib/lib/params/EventClassSpecParams.py
```diff
@@ -9,6 +9,7 @@
 
 from .OrganizerSpecParams import OrganizerSpecParams
 from .EventClassMappingSpecParams import EventClassMappingSpecParams
+from .SpecParams import SpecParams
```
cluther: I don’t see why `SpecParams.__init__(self)` is called in those cases instead of `OrganizerSpecParams.__init__(self)`. Going to the grandparent instead of parent seems questionable.

PR #479, opened 2017-09-11
Fix slow component grid loading

j053ph4:
Revise 'dataPointsToFetch' to include only metrics designated for use in
component grid columns.  Previous behavior loaded all metrics for all
components on a given device, which resulted in the slow loading
behavior observed

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/ClassSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/ClassSpec.py
@@ -1121,13 +1121,7 @@ class ClassSpec(Spec):
 
     def get_datapoints_to_fetch(self):
         """return list of datapoint names for templates associated with this class"""
-        dsnames = []
-        for dc_spec in self.zenpack.device_classes.values():
-            for t_name, t_spec in dc_spec.templates.items():
-                if t_name in self.monitoring_templates:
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/ZenPackLib/lib/spec/ClassSpec.py
```diff
@@ -1121,13 +1121,7 @@ def datapoints_to_fetch(self):
...
     def get_datapoints_to_fetch(self):
         """return list of datapoint names for templates associated with this class"""
-        dsnames = []
-        for dc_spec in self.zenpack.device_classes.values():
-            for t_name, t_spec in dc_spec.templates.items():
-                if t_name in self.monitoring_templates:
-                    dsnames += t_spec.datapoints_to_fetch
-        dsnames = list(set(dsnames))
-        return dsnames
+        return [spec.datapoint for spec in self.properties.values() if spec.datapoint and spec.grid_display]
```
cluther: Looks good, but I believe this makes RRDTemplateSpec.datapoints_to_fetch and RRDTemplateSpec.get_dp_names dead code that should be cleaned up along with this change.

j053ph4: OK the dead code has been removed from RRDTemplateSpec as well

PR #483, opened 2017-10-18
add "reset" option to DeviceClassSpec (ZPS-810)

j053ph4:
- Added optional "reset" parameter to facilitate override of zProperties
on preexisting device class
- Added test_device_class_existing_zproperties unit test


joshw:
I think this needs a doc update as well.  This seems like a feature that we would only rarely want to use, so it's important to point out why it's potentially a bad idea to use it, and where it is appropriate.

j053ph4:
I updated the docs per your suggestion along with a cautionary tale about bad stuff that could happen.  I only listed one scenario while indicating others exist and that migrate scripts are the way to go

j053ph4:
Thanks Josh!

PR #487, opened 2017-10-26
Fix failing unit test for impact triggers (ZPS-2271)

j053ph4:
- Fixes ZPS-2271
- Rewrote test_impact_triggers to fix failure
- Also removed implicit zope connection that can cause build failures

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
+++ b/ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
@@ -132,13 +132,14 @@ class ZPLTestMockZenPack(BaseTestCase):
     yaml_doc = '''name: ZenPacks.zenoss.ZenPackLib'''
     zenpack_name = 'ZenPacks.zenoss.MockZenPack'
     disableLogging = True
-    datsources = []
-    thresholds = []
 
-    def afterSetUp(self):
+    def afterSetUp(self, datasources=None, thresholds=None):
         super(ZPLTestMockZenPack, self).afterSetUp()
-
-        self.zenpack = MockZenPack(self.dmd, self.zenpack_name, datasources=self.datasources, thresholds=self.thresholds)
+        if datasources is None:
+            datasources = []
+        if thresholds is None:
+            thresholds = []
+        self.zenpack = MockZenPack(self.dmd, self.zenpack_name, datasources=datasources, thresholds=thresholds)
         # for a list of docs, iterate through
         if isinstance(self.yaml_doc, list):
             counter = 0
@@ -163,9 +164,13 @@ class ZPLTestMockZenPack(BaseTestCase):
 
 class MockZenPack(object):
     ''''''
-    def __init__(self, dmd, name, datasources=[], thresholds=[]):
+    def __init__(self, dmd, name, datasources=None, thresholds=None):
         self.dmd = dmd
         self.name = name
+        if datasources is None:
+            datasources = []
+        if thresholds is None:
+            thresholds = []
         self.datasources = datasources
         self.thresholds = thresholds
         self.install_zenpack()
--- a/ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py
+++ b/ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py
@@ -14,14 +14,9 @@ import Globals  # noqa
 from Products.ZenUtils.Utils import unused
 unused(Globals)
 
-from zope.component import getGlobalSiteManager
-from Products.ZenModel.Device import Device
 from Products.ZenUtils.guid.interfaces import IGlobalIdentifier
-from Products.ZenUtils.guid.interfaces import IGloballyIdentifiable
-# from zenoss.protocols.protobufs import model_pb2 as MODELCONSTANTS
 from Products.ZenMessaging.queuemessaging.publisher import ModelChangePublisher
 from zenoss.protocols.protobufutil import ProtobufEnum
-
 from ZenPacks.zenoss.ZenPackLib.lib.utils import impact_installed
 
 class ImpactTestCase(object):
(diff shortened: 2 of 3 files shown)


j053ph4:
bump please

ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py
```diff
@@ -77,23 +67,30 @@ class ImpactTestCase(object):
...
+  /Server/Basic:
+    remove: true
+    zProperties:
+      zPythonClass: ZenPacks.zenoss.MockZenPack.BasicDevice
+
 """
 
 
-class TestImpactTriggers(ImpactTestCase):
+class TestImpactTriggers(ImpactTestCase, ZPLTestMockZenPack):
```
cluther: What is the purpose of ImpactTestCase? It appears to be an empty class extending only `object` to me. Why have it at all?

ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py
```diff
@@ -77,23 +67,30 @@ class ImpactTestCase(object):
...
     """Test Impact Triggers"""
 
+    datasources = []
```
cluther: Class level properties should not be set to mutable values like this. It is a common source of bugs. This goes for thresholds on the following line too. It also goes for the same properties on the ZPLTestMockZenPack class.

The same is also true for method/function default values in `MockZenPack.__init__(self, dmd, name, datasources=[], thresholds=[])`.

ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py
```diff
@@ -77,23 +67,30 @@ class ImpactTestCase(object):
...
+    thresholds = []
+    yaml_doc = YAML_DOC
+    zenpack_name = 'ZenPacks.zenoss.MockZenPack'
+
     def afterSetUp(self):
         super(TestImpactTriggers, self).afterSetUp()
-        self.z = ZPLTestHarness(YAML_DOC)
-        self.z.connect()
+        super(ZPLTestMockZenPack, self).afterSetUp()
```
cluther: This isn't appropriate usage of `super`. The first argument to super should also be the current class, not one of its parent classes. To invoke a method in a specific class you should use `ZPLTestMockZenPack.afterSetUp(self)`. Though in this case you should probably get rid of the useless `ImpactTestCase` parent class and use super the normal way since `ZPLTestMockZenPack` would be the only parent class.

ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
```diff
@@ -132,13 +132,11 @@ class ZPLTestMockZenPack(BaseTestCase):
     yaml_doc = '''name: ZenPacks.zenoss.ZenPackLib'''
     zenpack_name = 'ZenPacks.zenoss.MockZenPack'
     disableLogging = True
-    datsources = []
-    thresholds = []
 
-    def afterSetUp(self):
+    def afterSetUp(self, datasources=[], thresholds=[]):
```
cluther: Mutable default arguments are just as bad a mutable class-level property values. You should do this instead.

```
def afterSetUp(self, datasources=None, thresholds=None):
    if datasources is None:
        datasources = []

    if thresholds is None:
        thresholds = []
```

ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py
```diff
@@ -14,38 +14,29 @@
...
-from Products.ZenModel.Device import Device
 from Products.ZenUtils.guid.interfaces import IGlobalIdentifier
-from Products.ZenUtils.guid.interfaces import IGloballyIdentifiable
-# from zenoss.protocols.protobufs import model_pb2 as MODELCONSTANTS
 from Products.ZenMessaging.queuemessaging.publisher import ModelChangePublisher
 from zenoss.protocols.protobufutil import ProtobufEnum
-
 from ZenPacks.zenoss.ZenPackLib.lib.utils import impact_installed
 
 class ImpactTestCase(object):
```
cluther: Can this `ImpactTestCase` class be removed now?

ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
cluther: Would you also fix `MockZenPack.__init__` in the same way?

ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py
```diff
@@ -77,23 +68,28 @@ class ImpactTestCase(object):
...
 
-class TestImpactTriggers(ImpactTestCase):
+class TestImpactTriggers(ZPLTestMockZenPack):
     """Test Impact Triggers"""
 
+    yaml_doc = YAML_DOC
+    zenpack_name = 'ZenPacks.zenoss.MockZenPack'
+
     def afterSetUp(self):
         super(TestImpactTriggers, self).afterSetUp()
```
cluther: Getting the following test failure on this now.

```
Error in test testFillTriggerProtoBuf (ZenPacks.zenoss.ZenPackLib.tests.test_impact_triggers.TestImpactTriggers)
Traceback (most recent call last):
  File "/usr/lib64/python2.7/unittest/case.py", line 360, in run
    self.setUp()
  File "/opt/zenoss/lib/python2.7/site-packages/Testing/ZopeTestCase/base.py", line 91, in setUp
    self.afterSetUp()
  File "/mnt/src/ZenPacks.zenoss.ZenPackLib/ZenPacks/zenoss/ZenPackLib/tests/test_impact_triggers.py", line 87, in afterSetUp
    super(TestImpactTriggers, self).afterSetUp()
TypeError: afterSetUp() takes exactly 3 arguments (1 given)


  Ran 1 tests with 0 failures and 1 errors in 10.772 seconds.
Tearing down left over layers:
  Tear down Products.ZenTestCase.BaseTestCase.ZenossTestCaseLayer in 0.000 seconds.
  Tear down Testing.ZopeTestCase.layer.ZopeLite in 0.000 seconds.

Tests with errors:
   testFillTriggerProtoBuf (ZenPacks.zenoss.ZenPackLib.tests.test_impact_triggers.TestImpactTriggers)
```

PR #492, opened 2017-11-08
Update test_proxy_path_reporters

j053ph4:
Use IpService proxy class instead of HWProduct since platform has
deprecated the productClass relation


j053ph4:
Thanks Phil

PR #494, opened 2017-11-08
Truncate ZP install logging regarding template changes (ZPS-1520)

j053ph4:
- Fixes ZPS-1520
- remove YAML diff from logging output, and corrected initial reference
to template name in log message


j053ph4:
tested with product-assembly as well

PR #496, opened 2017-11-09
Detect and backup user-performed template changes (ZPS-1520)

j053ph4:
- Fixes ZPS-1520
- Moves change detection and backup behavior to ZenPack.remove() method
- Effectively restricts change detection, reporting, and backup to
user-performed changes (when upgrading a ZPL2-based ZP)
- Upgrades from ZPL1 use the ZPL1 remove method (which performs
similarly)
- Upgrades from pre-ZPL ZenPacks perform no change detection or backup
of user templates.

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
@@ -70,12 +70,10 @@ class ZenPack(ZenPackBase):
 
         # load monitoring templates
         for dcname, dcspec in self.device_classes.iteritems():
-            dcspecparam = self._v_specparams.device_classes.get(dcname)
             deviceclass = dcspec.get_organizer(app.zport.dmd)
 
             for mtname, mtspec in dcspec.templates.iteritems():
-                mtspecparam = dcspecparam.templates.get(mtname)
-                self.update_object(app, deviceclass, 'rrdTemplates', mtname, mtspec, mtspecparam)
+                mtspec.create(app.zport.dmd)
 
         # Load event classes
         for ecname, ecspec in self.event_classes.iteritems():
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -109,15 +104,16 @@ def remove(self, app, leaveObjects=False):
...
-                    # back it up if it exists
                     if template:
-                        self.get_or_create_backup(deviceclass, 'rrdTemplates', orig_mtname)
+                        self.backup_user_changes(app, deviceclass, 'rrdTemplates',
+                                                 orig_mtname, orig_mtspec, orig_mtspecparam)
                     else:
                         self.LOG.warning(
-                            "Monitoring template {}/{} has been removed at some point "
-                            "after the {} ZenPack was installed.  It will be "
+                            "Monitoring template {}/{} has either been removed or is new to this version "
```
joshw: I think "new to this version" is no longer applicable if I understand the change- this is at remove time now, so it's only dealing with the currently installed vs. distributed

ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -168,62 +164,30 @@ def remove(self, app, leaveObjects=False):
...
-            try:
-                return rel._getOb(object_id)
-            except AttributeError:
-                pass
-        return None
+        # if they differ, create a backup
+        if diff:
+            backup = self.get_or_create_backup(parent, relname, object_id)
+            LOG.info("Existing object {}/{} differs from "
+                     "the newer version included with the {} ZenPack.  "
```
joshw: Same comment here

PR #498, opened 2017-11-20
Handle traceback if threshold ZP not installed

yichi-lu:
Fixes ZPS-1732

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/spec/RRDTemplateSpec.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/spec/RRDTemplateSpec.py
@@ -138,7 +138,10 @@ class RRDTemplateSpec(Spec):
 
         self.speclog.debug("adding {} thresholds".format(len(self.thresholds)))
         for threshold_id, threshold_spec in self.thresholds.items():
-            threshold_spec.create(self, template)
+            try:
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/ZenPackLib/lib/spec/RRDTemplateSpec.py
```diff
@@ -138,7 +138,10 @@ def create(self, dmd, addToZenPack=True, id=None):
 
         self.speclog.debug("adding {} thresholds".format(len(self.thresholds)))
         for threshold_id, threshold_spec in self.thresholds.items():
-            threshold_spec.create(self, template)
+            try:
+                threshold_spec.create(self, template)
+            except ValueError as err:
+                self.LOG.error("threshold creation error: %s", err.message)
```
cluther: I'm not sure about this. I think the point of marking a threshold `optional: false` in the YAML is to make the ZenPack fail to install if the threshold type doesn't exist on the system. It'd be nice if we could somehow fail the ZenPack installation without a traceback, but I don't know that we have a way to do that.

Ideally any ZenPack that specified an `optional: false` threshold in its YAML would also include a requirement on the ZenPack that provides that threshold type (i.e. Capacity, DurationThreshold, etc.) in its setup.py.

PR #501, opened 2018-02-12
Make most unit tests optional (ZPS-3035)

j053ph4:
- Fixes ZPS-3035
- This is a temporary change until ZPLTestBase and associated classes
are redesigned to avoid automated build issues (ZPS-3035)
- Unit tests are skipped unless the "RUN_ZPL_TESTS" environment variable
is set beforehand
- Updated __version__ in zenpacklib.py to 2.0.10

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
+++ b/ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
@@ -30,6 +30,12 @@ from ZenPacks.zenoss.ZenPackLib.tests.ZPLTestHarness import ZPLTestHarness
 from ZenPacks.zenoss.ZenPackLib.lib.helpers.loaders import WarningLoader
 from ZenPacks.zenoss.ZenPackLib.lib.helpers.ZenPackLibLog import ZPLOG
 
+RUN_TESTS = False
+if os.getenv('RUN_BROKEN_TESTS'):
+    RUN_TESTS = True
+else:
+    print "Set RUN_BROKEN_TESTS environment variable to enable unit testing"
+
 
 class ZPLTestBase(BaseTestCase):
     """BaseTestCase class with ZPLTestHarness support"""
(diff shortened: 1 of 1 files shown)


ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
```diff
@@ -30,6 +30,12 @@
 from ZenPacks.zenoss.ZenPackLib.lib.helpers.loaders import WarningLoader
 from ZenPacks.zenoss.ZenPackLib.lib.helpers.ZenPackLibLog import ZPLOG
 
+run_tests = False
```
cluther: Globals should be uppercase: `RUN_TESTS`.

Also I think the name and the name of its corresponding environment variable are a bit of a misnomer. It would make me think that it enables/disables all ZPL tests. Maybe broken is a harsh word, but something like `RUN_BROKEN_TESTS` or `RUN_FLAKY_TESTS` would convey the meaning better.

ZenPacks/zenoss/ZenPackLib/tests/ZPLTestBase.py
```diff
@@ -38,6 +44,13 @@ class ZPLTestBase(BaseTestCase):
     disableLogging = True
     z = None
 
+    def setUp(self):
+        if not run_tests:
+            msg = 'Skipping {} because environment '\
```
cluther: Just a style thing: try to avoid backslash line-continuations. Especially this case where the continuation isn't even indented. I think something like the following reads better, and style checkers are usually happier with.

```python
msg = (
    "Skipping {} because environment "
    "variable RUN_ZPL_TESTS is unset."
    ).format(self.__class__.__name__)
```

ZenPacks/zenoss/ZenPackLib/zenpacklib.py
```diff
@@ -27,7 +27,7 @@
 """
 
 # PEP-396 version. (https://www.python.org/dev/peps/pep-0396/)
-__version__ = "2.0.9"
+__version__ = "2.0.10"
```
cluther: Version bump should be made directly on the hotfix/2.0.10 branch before this commit.

PR #503, opened 2018-03-05
Add default "capacity_get_threshold_configs" method to ZenPack.py

j053ph4:
- This should facilitate automatic generation of Capacity ZP thresholds
for the case when Capacity ZP is installed after a given ZP with
Capacity thresholds defined.


j053ph4:
I added documentation and tests, and also updated this fix to automatically handle both installed and not installed scenarios automatically (meaning that the capacity thresholds can now be included in the YAML file without the earlier workaround)

j053ph4:
Trotting out the slippery slope argument here...why stop at thresholds then?  Why not deal with missing datasource or datapoint or graphpoint classes as well?  Would that then obligate us to come up with various workarounds for non-sequential installation/dependency satisfaction?

j053ph4:
Also I wanted to point out that the "optional" flag for thresholds exists for ThresholdSpec as well

PR #511, opened 2018-04-25
Ensure that zenpacklib exports YAML as expected (ZPS-1746)

j053ph4:
- Fixes ZPS-1746
- Updated EventClassSpec and ProcessClassOrganizer spec
create/instantiate methods to return organizers
- Added missing 'replaceRegex' attribute to ProcessClassSpecParams
- Moved zenpack dumping methods to YAMLExport class mixin, where they
can be unit tested without calling ZenScriptBase or the command line
- Updated ZPLCommand to use YAMLExport methods for dumping
- Added test_yaml_export.py unit test

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/helpers/YAMLExporter.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/helpers/YAMLExporter.py
@@ -1,0 +1,143 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2016, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+import yaml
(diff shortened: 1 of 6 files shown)


cluther:
I’m not sure why ZPS-1746 is even a problem. I thought we had decided that objects added by zenpacklib from YAML wouldn’t be exported. The only process classes that should be exported by `zenpacklib --dump-process-classes` are those that were loaded from XML, or created by the user and associated with a ZenPack.

This is related to ZPS-1601, and subsequent ZPS-2660 that had to fix the regression its fix caused.

There’s no need to export YAML for something when you can just find the YAML in the source file already.

I think the test case for ZPS-1746 is wrong. We should be testing by manually creating the process class through the UI, adding it to the ZenPack, then using `zenpacklib --dump-process-classes` to see if it dumps to YAML.

PR #521, opened 2018-05-16
Support for custom datapoint classes

j053ph4:
- Remove need to override manage_addRRDDataPoint method
- Added "type" attribute to RRDDatapointSpec.  If unspecified, calls
ds.manage_addRRDDataPoint (backwards compat).  If specified, creates the
datapoint directly and adds to ds datapoints.
- Added "_getClassesByPath" ZP method override for unit testing
- Added custom datapoint testing to test_extra_params_types.py


csmouse:
PR has aged out. Closing.

PR #543, opened 2018-08-09
ZEN-29068: For upgrades, a separate log file needs to be generated fo…

olst:


olst:
https://jira.zenoss.com/browse/ZEN-29068?focusedCommentId=134236&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-134236

j053ph4:
Another thought -- if this is a particular pain point for vSphere ZP, what about making these changes in a ZenPack class override in the vSphere ZP __init__ (ZenPack) class?

olst:
The request was to save __all__ templates' diffs into a separate place. This is not related only to vSphere ZP.

olst:
please see https://github.com/zenoss/ZenPacks.zenoss.ZenPackLib/pull/544

dougsyer:
Would be nice to be able to duel log all zplib commamds to a separate log thst is set to go to kibana.  Also be able to have it use a different level of severity than what goes to stdout.

When i install a zenpack i just want to know if tgerr are errors amd it should be essy to see then refer to a detailed log for details om success or failure.

Also there is no clesr indicator that a zp sucrssfully installed and the errored output usually makes it clear something went wrong but its inconsistent so i find auromating it with a lipeline and checking for failures to be difficult

PR #551, opened 2019-04-23
ZEN-28879 Critical event on /Status shows device as down, should only…

Merteg:
… show as down for /Status/Ping


Merteg:
Main PR: https://github.com/zenoss/zenoss-prodbin/pull/3593

PR #553, opened 2019-08-23
ZEN-32176 Products.Zing tests are failing

Merteg:
After some tests, that inherited on ZPLTestCaseLayerBase, Transaction.abort method still overridden by empty lambda. Need to turn it back before test finish


Merteg:
Wrong base branch

PR #558, opened 2019-11-22
REVIEW ONLY DO NOT MERGE

j053ph4:


ZenPacks/zenoss/ZenPackLib/lib/spec/RRDTemplateSpec.py
```diff
@@ -166,3 +167,10 @@ def remove(self, dmd, id=None):
         existing_template = device_class.rrdTemplates._getOb(t_id, None)
         if existing_template:
             device_class.rrdTemplates._delObject(t_id)
+
+    def getDataPointClasses(self, dmd):
```
StevePC: might be worthwhile to memoize this type of call so we don't need to walk all the ZenPacks for the same set of results every time.

PR #560, opened 2020-02-25
Prepare code for upgrade PyYAML to 5.x version

DeKapito:
Fixes ZEN-31325

Fix tests and replace yaml.Loader with yaml.FullLoader (FullLoader is
the safe version of Loader. Avoids arbitrary code execution). This
commit is changing code to be compatible with PyYAML 5.x and saving
backwards compatibility

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
@@ -256,13 +256,13 @@ class ZenPack(ZenPackBase):
     def object_changed(self, app, object, spec, specparam):
         """Compare new and old objects with prototype creation"""
         # get YAML representation of object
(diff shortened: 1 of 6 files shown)


ZenPacks/zenoss/ZenPackLib/lib/base/ZenPack.py
```diff
@@ -256,13 +256,13 @@ def move_object(self, parent, relname, source_id, dest_id):
     def object_changed(self, app, object, spec, specparam):
         """Compare new and old objects with prototype creation"""
         # get YAML representation of object
-        object_yaml = yaml.dump(specparam.fromObject(object), Dumper=Dumper)
+        object_yaml = yaml.dump(specparam.fromObject(object), Dumper=Dumper, default_flow_style=None)
```
j053ph4: what effect does this have on comparing diffs?

PR #563, opened 2020-04-08
Adding model relationships containment check

jeff-powell:
This is a generic test, to be implemented within target zenpacks, which validates the model containments.  It verifies that each component is only contained once, otherwise the unit tests assertion fails.  https://jira.zenoss.com/browse/ZPS-6322

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/tests/ModelContainmentTestCase.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/tests/ModelContainmentTestCase.py
@@ -1,0 +1,92 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2020, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+import importlib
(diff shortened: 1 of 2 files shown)


ZenPacks/zenoss/ZenPackLib/lib/tests/ModelContainmentTestCase.py
```diff
@@ -0,0 +1,92 @@
...
+        class_relationships = self.get_relationship_maps()
+
+        # Iterate Class Relationships
+        self.assertIsNotNone(class_relationships)
+        self.assertTrue(isinstance(class_relationships, list))
+        containment_list = []
+        for rel in class_relationships:
+            name = getattr(rel, 'left_relname', '')
+            type = getattr(rel, 'left_type', '')
+            if type == 'ToManyCont':
```
StevePC, line 90: while an edge case for sure, should this also check for ToOneCont?

jeff-powell: that can be added.  (I don't see 'ToOneCont' very often)  The same principle applies, that each component should only be under a single containment hierarchy.

PR #566, opened 2023-10-20
ZEN-34366: ZPS-8617: don't try to remove replaced template if not there

StevePC:


odubrovyk:
@StevePC I noticed that the previous ZenPackLib build failed after the template-related changes. The reason is failed unit tests - http://solutions-jenkins.zenoss.eng/job/ZenPackLib-RunTests/110/console. So, I suspect that the build with these changes will also fail. Could you please take a look and correct the failed unit tests?

odubrovyk (approved):

PR #567, opened 2023-12-12
Feature: ZEN-34366

StevePC:


odubrovyk:
I reviewed your comment in Jira and this fix. It is clear to me. Could you please change a destination branch to a hotfix/2.1.3 branch? We are planning the next release for ZPL as a hotfix. So, all current and future changes should be targeted into the hotfix/2.1.3 branch.

PR #570, opened 2023-12-21
Feature/ZEN-34366 migration for replacement/addition changes to zDeviceThresholds

StevePC:


ZenPacks/zenoss/ZenPackLib/migrate/FixReplacementTemplates.py
```diff
@@ -0,0 +1,62 @@
...
+
+class FixReplacementTemplates(ZenPackMigration):
+    """
+    Change from AristaBgpPeer plugin to AristaVrfBgpMap plugin
+    """
+
+    version = Version(1, 6, 0)
+
+    def _migrateObject(self, obj):
+            if obj.hasProperty('zDeviceTemplates'):
```
odubrovyk, line 27: Probably won't affect anything, but I see a bigger indent here than it should be.

odubrovyk (changes requested):
@StevePC I left a couple of comments for this PR. 
I see that ZPS-8777 is not related to this issue. So, I think that a separate ZPS issue should be created for this fix and the following templates-related fixes. We can then link the new ZPS issue to ZEN-34366. Also, the description of the PR points to zDeviceThresholds, but the migration provided for zDeviceTemplates.

StevePC:
wrong jira again, fixing it this time.

PR #575, opened 2024-01-09
ZPS-8777: when not installing ZP, don't load yaml files that are only…

StevePC:
… needed at install time

### DIFF
--- a/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
+++ b/ZenPacks/zenoss/ZenPackLib/lib/helpers/utils.py
@@ -7,6 +7,7 @@
 #
 ##############################################################################
 import os
+import sys
 from collections import OrderedDict, Mapping
 import yaml
 import time
@@ -15,6 +16,8 @@ from .Dumper import Dumper
 from .loaders import OrderedLoader, ZenPackSpecLoader, ZenPackSpecFromDictLoader
 from ..base.ZenPack import ZenPack
 import inspect
+import re
+from StringIO import StringIO
 
 # list of yaml sections with DEFAULTS capability
 YAML_HAS_DEFAULTS = ['classes', 'properties', 'thresholds', 'datasources',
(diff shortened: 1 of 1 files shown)


joshw:
I generally like the idea behind this change, but I think it would break something I have been planning to merge (https://github.com/zenoss/zenoss-prodbin/commit/08dc609261fe9f3823bb75f1717cfddda760f8fa#diff-69550b8b649cc02b3f01c385454d956839461dbf9d8543953c6f03a67e2307f2R111) - basically, I want to figure out what zenpack "owns" a device, and part of that involves looking at the device classes that contain it to see what zenpack created them.  Right now, for ZPL zenpacks, that means looking at pack.device_classes, which i think your change might cause to be unpopulated.   It doesn't seem simple to work around this either.

StevePC:
> I generally like the idea behind this change, but I think it would break something I have been planning to merge ([zenoss/zenoss-prodbin@08dc609#diff-69550b8b649cc02b3f01c385454d956839461dbf9d8543953c6f03a67e2307f2R111](https://github.com/zenoss/zenoss-prodbin/commit/08dc609261fe9f3823bb75f1717cfddda760f8fa#diff-69550b8b649cc02b3f01c385454d956839461dbf9d8543953c6f03a67e2307f2R111)) - basically, I want to figure out what zenpack "owns" a device, and part of that involves looking at the device classes that contain it to see what zenpack created them. Right now, for ZPL zenpacks, that means looking at pack.device_classes, which i think your change might cause to be unpopulated. It doesn't seem simple to work around this either.

Typically we've looked at the class of the device and traced that back to the zenpack, which should always be there.  In reality, since the customer can create deviceclasses and add any kind of device to them, trying to trace device ownership by deviceclass doesn't make much sense.

odubrovyk (approved):
@StevePC Approved. In case you plan additional changes in this scope you can open a separate Pull Request. QAs need the fix in the migration file to test previous changes. So, you can merge this one.

PR #578, opened 2024-04-10
Fix ZP removal if suborganizer does not exist

odubrovyk:
Fixes ZPS-8863.

Skip the suborganizer removing if it has been removed previously.


odubrovyk:
@StevePC Could you please take a look at this PR?

StevePC (approved):

